PR #11, opened 2015-05-11
Tracks Client crash on iOS 7 with unrecognized selector

### LINKED ISSUE #10: Crash on iOS 7 in TracksEvent validation
Crash with the following code as containsString isn't available in iOS 7:

`TracksEvent.m`
```
    } else if ([(NSString *)*ioValue containsString:@"-"]) {
```

props to @koke

astralbodies:
Fixes #10 

Use a method available in iOS 7 to determine if a string contains a dash character. Also updated the deployment target to iOS 7 from base (currently 8.2) to allow unit tests to find these problems in the future.


astralbodies:
Merging as I need to get this out :cry:

koke:
Late, but :shipit:

PR #17, opened 2015-06-09
Wrap all public persistence methods in performBlockAndWait

### LINKED ISSUE #16: Wrong thread when fetching on CoreData
On this line https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Automattic-Tracks-iOS/TracksEventPersistenceService.m#L38

shouldn't be called a performBlockAndWait on the MOC?

astralbodies:
Fixes #16 

Wrap all persistence methods with `performBlockAndWait:` to make sure they're executed with some thread safety. This isn't a huge deal for WPiOS since the tracker thread is on its own thread but it prevents problems.

Needs Review: @SergioEstevao

### DIFF
--- a/Automattic-Tracks-iOS/TracksEventPersistenceService.m
+++ b/Automattic-Tracks-iOS/TracksEventPersistenceService.m
@@ -32,21 +32,25 @@
 
 - (NSArray *)fetchAllTracksEvents
 {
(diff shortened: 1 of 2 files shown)


SergioEstevao:
:shipit:

PR #23, opened 2015-08-03
Use a constant for setting the podspec version

### LINKED ISSUE #20: Automatically pull in client version number in events instead of fixed string
Refers to: https://github.com/Automattic/Automattic-Tracks-iOS/blob/b8dad1bdc8c417217bb29a7ec8ebe1d637c7f1a7/Automattic-Tracks-iOS/TracksService.m#L248

Store the current version of the client in the project and pull it out somehow. Because this project is being consumed as a CocoaPod, figure out the right way to persist the version number to be usable in the host project. CocoaPods might provide a macro we can use.

astralbodies:
Closes #20 

Establishes a version number for the library in TracksConstants.h/m which can be used in `TracksDeviceInformation` when sending events into the Nosara Tracks system. This same version number is now pulled into the podspec so the version number lives only in one place. Updating the version forces the constant to be updated.

Needs Review: @daniloercoli, @jleandroperez 

Thanks to @orta for the help coming up with this idea!

### DIFF
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -13,5 +13,5 @@ Pod::Spec.new do |spec|
   spec.framework    = 'CoreTelephony'
 
   spec.dependency 'UIDeviceIdentifier', '~> 0.4'
-  spec.dependency 'CocoaLumberjack', '~> 2.0'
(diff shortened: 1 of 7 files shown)


jleandroperez:
Nice trick!  :shipit:

PR #24, opened 2015-08-05
Adds Reachability for resilience & WiFi detection

### LINKED ISSUE #2: Incorporate Reachability into project
Use Reachability to disable timer & provide WiFi indicator in DeviceInformation.

astralbodies:
Closes #2 

Reachability has been added to help with two specific tasks:
1. Detect when WiFi is in use for the `device_info_wifi_connected` property. Up until now every event has been set to NO.
2. Disable the timer to send events when the network is detected as being offline. This helps reduce potential battery drain continually attempting to hit the host.

**Testing:**
Use the TracksDemo application. The easiest way to test this is probably by adding an `NSLog` statement to `StatsServiceRemote` to dump out the `dataToSend` contents. You can then verify the right flags are attached to the events being sent to Tracks. Possible scenarios:

1. Turn on WiFi, record an event.
2. Turn off WiFi (wait like 2 secs for Reachability to detect) and record an event.
3. Turn off cellular + WiFi, record an event.
4. Turn off all data (airplane mode), record an event.

Depending on the device you're testing and if you have cellular and WiFi or just WiFi these scenarios will be a little different. Here is the expected outcome:

1. No data at all - Events will get recorded but WiFi will be listed as NO and the remote call & timer should never fire.
2. Only cellular - Events will get recorded, WiFi is listed as NO. Remote call should still happen, timer remains on and firing every 10sec.
3. Only WiFi - Events will get recorded, WiFi is listed as YES. Remote call should still happen, timer remains on and firing every 10sec.
4. Both cellular + WiFi - Events will get recorded, WiFi is listed as YES. Remote call should still happen, timer remains on and firing every 10sec.
5. App background - events should stop with one last attempt immediately upon backgrounding.

Needs Review: @jleandroperez 
cc: @daniloercoli


Automattic-Tracks-iOS/TracksService.m
```diff
@@ -199,6 +219,31 @@ - (void)didBecomeActive:(NSNotification *)notification
...
+    Reachability *reachability = (Reachability *)notification.object;
+    
+    // Because the containing app may already use Reachability, limit this to ours only.
+    if (reachability != self.reachability) {
+        return;
+    }
+
+    self.deviceInformation.isWiFiConnected = reachability.isReachableViaWiFi;
+    
+    if (reachability.isReachable == YES && self.isHostReachable == NO) {
```
jleandroperez: Please, *beware*: we had [a severe bug in Simperium](https://github.com/Simperium/simperium-ios/issues/287), which took blood to catch with the debugger, and caused lots of Out of Sync scenarios.

Reachability, when initialized to check networking against a specific domain, is not reliable (there was even a demo somewhere showing up corner cases, with Apple's own sample code).

Their recommendation (WWDC's labs) was to [initialize for Internet Connection check](https://github.com/Simperium/simperium-ios/blob/develop/Simperium/Simperium.m#L84), instead of host checks, and... just [verify the status of the currentReachability](https://github.com/Simperium/simperium-ios/blob/develop/Simperium/Simperium.m#L233) flag.

astralbodies:
For now I'm going to leave it as it is with the hostname check as I need WiFi detection. Doesn't look like WiFi detection works unless you're checking a hostname or address.

When the app is backgrounded, I do stop the notifier and force one more send of events to Tracks. When the app is foregrounded the notifier is started again. I'll keep an eye on things in development.

jleandroperez:
@astralbodies Acknowledged! if anything, please, just let me know!

PR #25, opened 2015-08-05
Sets user agent on events properly and also now on URL connection

### LINKED ISSUE #21: Evaluate User Agent
```
Android useragent: Nosara Client for Android/1.0.3
iOS useragent: WordPress/5.3.1 CFNetwork/711.4.6 Darwin/14.0.0
```

Make sure we're setting the user agent on connections make to the Tracks endpoint.

astralbodies:
Closes #21 

Sets the user agent on all of the tracks events. It was not being set but the functionality was there - it was an incomplete feature. The current user agent is still being used in the parent property calculation.

The user agent is also now being set on the NSURLConnection used to connect to the REST endpoint.

Not entirely sure what the best approach is to test this other than reviewing the events sent to Tracks or inspecting with Charles Proxy.

Needs Review: @aerych, @daniloercoli

### DIFF
--- a/Automattic-Tracks-iOS/TracksService.m
+++ b/Automattic-Tracks-iOS/TracksService.m
@@ -8,6 +8,7 @@
 @property (nonatomic, assign) BOOL timerEnabled;
 @property (nonatomic, strong) TracksContextManager *contextManager;
 
+@property (nonatomic, readonly) NSString *userAgent;
 @property (nonatomic, copy) NSString *username;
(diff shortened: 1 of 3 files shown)


aerych:
Tested with Charles and everything looked good. 
:shipit: from me. :)

astralbodies:
Your time is definitely appreciated @aerych! @daniloercoli we discussed this in Slack so I think we're good.

PR #29, opened 2015-08-06
Prevent long variable memory problems on 32-bit processors

### LINKED ISSUE #28: 32-bit devices getting strange timestamps due to memory limit
Calculate the timestamps in such a way that it doesn't require multiplying the NSTimeInternal by 1000 to get milliseconds. The number for time since 1970 is bigger than a 32-bit long. Events are getting rejected because the value isn't logical.

cc: @daniloercoli

astralbodies:
Closes #28 

On 32-bit devices it seems that our calculation for milliseconds since 1970 is too long to fit into memory. Instead of multiplying the double by 1000 to get a long, format the number and pad with three zeros.

Needs Review: @jleandroperez


jleandroperez:
Looks great! :shipit:

astralbodies:
@jleandroperez Can you review again? `lround` was the culprit.

jleandroperez:
Looking good. Way within `long long` range!

:shipit:

PR #39, opened 2015-10-13
Dynamic User Type

jleandroperez:
#### Description:
This Pull Request moves over internal private strings, used to indicate the User Type, to a public property.

This will allow us to update the Username Strings for Simplenote, so that we don't have an awful namespaces collision.

@astralbodies i'd be more than happy to change / update anything you say!.

Needs Review: @astralbodies 
Thanks in advance!

### DIFF
--- a/Automattic-Tracks-iOS/TracksService.h
+++ b/Automattic-Tracks-iOS/TracksService.h
@@ -12,6 +12,8 @@ extern NSString *const TrackServiceDidSendQueuedEventsNotification;
 @property (nonatomic, strong) TracksServiceRemote *remote;
 
 @property (nonatomic, copy) NSString *eventNamePrefix;
+@property (nonatomic, copy) NSString *anonymousUserType;
+@property (nonatomic, copy) NSString *authenticatedUserType;
 
 @property (nonatomic, assign) NSTimeInterval queueSendInterval;
 @property (nonatomic, readonly) NSUInteger queuedEventCount;
--- a/Automattic-Tracks-iOS/TracksService.h
+++ b/Automattic-Tracks-iOS/TracksService.h
@@ -12,8 +12,8 @@ extern NSString *const TrackServiceDidSendQueuedEventsNotification;
 @property (nonatomic, strong) TracksServiceRemote *remote;
 
 @property (nonatomic, copy) NSString *eventNamePrefix;
-@property (nonatomic, copy) NSString *anonymousUserType;
-@property (nonatomic, copy) NSString *authenticatedUserType;
+@property (nonatomic, copy) NSString *anonymousUserTypeKey;
+@property (nonatomic, copy) NSString *authenticatedUserTypeKey;
 
 @property (nonatomic, assign) NSTimeInterval queueSendInterval;
 @property (nonatomic, readonly) NSUInteger queuedEventCount;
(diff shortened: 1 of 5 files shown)


Automattic-Tracks-iOS/TracksService.m
```diff
@@ -116,7 +118,7 @@ - (void)trackEventName:(NSString *)eventName withCustomProperties:(NSDictionary
                                               username:self.username
                                                 userID:self.userID
                                              userAgent:self.userAgent
-                                              userType:self.isAnonymous ? TracksEventUserTypeAnonymous : TracksEventUserTypeWordPressCom
+                                              userType:self.isAnonymous ? TracksEventUserTypeAnonymous : TracksEventUserTypeAuthenticated
```
astralbodies: Should this be:
```objc
userType:self.isAnonymous ? self.anonymousUserType : self.authenticatedUserType;
```

Automattic-Tracks-iOS/TracksService.h
```diff
@@ -12,6 +12,8 @@ extern NSString *const TrackServiceDidSendQueuedEventsNotification;
 @property (nonatomic, strong) TracksServiceRemote *remote;
 
 @property (nonatomic, copy) NSString *eventNamePrefix;
+@property (nonatomic, copy) NSString *anonymousUserType;
+@property (nonatomic, copy) NSString *authenticatedUserType;
```
astralbodies: I think these property names aren't clear enough as to what they are. Maybe `anonymousUserTypeKey` and `authenticatedUserTypeKey`?

jleandroperez: Updating, thanks!

Automattic-Tracks-iOS/TracksService.m
jleandroperez: userType parameter is an NSInteger. That gets converted, down the line, into a NSString (by means of the new two properties!)

jleandroperez:
@astralbodies ready for a second look! thank you sir!

Automattic-Tracks-iOS/TracksService.m
astralbodies: Whoops I was getting confused with the string constants defined above. Derp.

astralbodies:
:shipit:

jleandroperez:
Thanks Aaron!

Automattic-Tracks-iOS/TracksService.m
```diff
@@ -54,7 +54,7 @@ @interface TracksService ()
 NSString *const TracksUsernameKey = @"_ul";
 
 NSString *const TracksUserTypeAnonymous = @"anon";
-NSString *const TracksUserTypeWPCOM = @"wpcom:user_id";
+NSString *const TracksUserTypeAuthenticated = @"wpcom:user_id";
```
daniloercoli: Is this new const string going to be used when tracking events for simplenote authenticated user? If so, there is a problem with it. We still need to decide which string will represent the Simplenote user on the server side.

astralbodies: The Simplenote app will set the `authenticatedUserTypeKey` parameter with something pre-determined, much like how `eventNamePrefix` works to override `wpios`.

PR #42, opened 2015-11-10
OS X Doesn't require CoreTelephony Framework

jleandroperez:
Tracks is crashing in Yosemite, since CoreTelephony isn't available there.
This wasn't picked up before, because CoreTelephony was added in El Capitan.

Needs Review: @astralbodies


astralbodies:
:shipit:

roundhill:
gojo_laptop

jleandroperez:
:smile:

PR #45, opened 2015-11-12
Fix data model loading when used as framework

koke:
I'm not sure if this line still needs to stay:

```obj-c
NSBundle *bundle = path != nil ? [NSBundle bundleWithPath:path] : [NSBundle bundleForClass:[self class]];
```

Needs Review: @astralbodies


koke:
Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/4434

astralbodies:
Nice! Thanks for the update @koke :smile: :shipit:

PR #53, opened 2016-07-21
Issue/52 use ephemeral session

### LINKED ISSUE #52: Use a dedicated ephemeral session instead of shared session.
Reasons:

 - Call to the tracks service don't need to cached in anyway by the network.
 - Avoid flooding the shared session with events.
 - Easier to cancel/interrupt sending of events in the future

SergioEstevao:
Closes #52

How to test:
 - Start the demo app
 - Send events and check if they get to the server correctly.

Needs Review: @astralbodies


astralbodies:
Seems to work well - :shipit:

PR #55, opened 2016-07-21
Validates HTTP response code & content

### LINKED ISSUE #54: Verify errored network response causes events to be re-queued
Make sure the mechanism in place for re-queuing events when a network call fails works properly.

astralbodies:
Fixes #54 

Turns out I was not verifying that the HTTP response code was valid - I had assumed NSURLSession turned a non-2xx response into an NSError. Oops.

This fixes that, adds an additional check looking for `"Success"` and unit tests to validate the changes.

Needs Review: @SergioEstevao


SergioEstevao:
:shipit:

PR #70, opened 2018-04-09
Prevent tracks sending events after disabled

### LINKED ISSUE #65: Retain cycles in tracks client
We've identified a couple of retain cycles which prevent the TracksService to be deallocated when it's not needed anymore. This has the side effect of sending queued events after the service is already released from the app.

Both the service and the remote reference `self` from a callback block without `__weak`:

screen shot 2017-07-07 at 11 17 29
screen shot 2017-07-07 at 11 19 40

koke:
This fixes a couple retain cycles that were preventing the Tracks service from being deallocated, so it was still sending events after the user had opted out. It also adds a safeguard to clear any queued events, in case there are more undetected leaks.

Fixes #65 

## To test

Sadly, I haven't been able to test that the leaks are fixed since Instruments keep crashing on me. You can check out the `issue/tracks-audit` branch in the WordPress app and make sure it's not sending events after analytics is disabled.

### DIFF
--- a/Automattic-Tracks-iOS/TracksEventPersistenceService.h
+++ b/Automattic-Tracks-iOS/TracksEventPersistenceService.h
@@ -14,6 +14,8 @@
 
 - (void)removeTracksEvents:(NSArray *)tracksEvents;
 
+- (void)clearTracksEvents;
+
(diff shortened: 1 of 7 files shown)


koke:
Thanks for the quick review 👌

jleandroperez:
Anytime sir!

Issue #63, opened 2017-03-14
Add Apple Watch paired to device_information


astralbodies:
Related: https://github.com/wordpress-mobile/WordPress-iOS/issues/6402

PR #76, opened 2018-11-23
Add support for apple watch detection

jkmassel:
Adds initial support for Apple Watch detection for Tracks events (which resolves https://github.com/Automattic/Automattic-Tracks-iOS/issues/63)

One gotcha – there's an initial delay while creating a watch connectivity session – because the process is asynchronous, it can be tricky to pull out the right information at any given moment.

One solution is to maintain a record of whether we've _ever_ had a watch connection on this device. Another bonus – having this record makes it so that we don't have to constantly spool up a watch connectivity session – it seems doing this actually does make contact with the watch (to determine whether an associated app is installed on the watch), which is bad for both device's batteries.

It'd be possible to have the result cached for a shorter period of time (rather than the "forever" timeframe it is right now) easily enough.

Consider this a starting point?

**To Test**
There's no way to test this in the simulator – this particular API seems to only work on-device. Additionally, it's not possible to run unit tests on-device, so I've whipped up a quick testing target in https://github.com/Automattic/Automattic-Tracks-iOS/tree/add/on-device-testing-target that you can run on-device. All it does is display `DeviceInformation` keys that can't be Unit Tested.


astralbodies:
@jkmassel I'll be reviewing this one later in the week; going to need a bit more time for testing on my side. 😄

astralbodies:
@jkmassel I suck at reviews apparently and remembering to do them.

Quick question - does this work for an app that doesn't have a watch extension yet?

jkmassel:
It seems to! If you switch to the `add/on-device-testing-target` branch, and run the `Automattic-Tracks-Testing` target on-device (with a paired watch), it'll work.

One caveat – you need to launch the app, quit, then launch it again. This is because of the async nature of the Watch Connectivity APIs. 

IMHO it's better to not block tracks calls until we hear back from the API, but instead, to just start updating events once we do – I'd imagine that we'll get enough data that way.

If that approach is no good though, let me know – happy to rework it!

astralbodies:
@jkmassel I think the approach is solid! I'll update the branch and when it goes green, go ahead ahead and merge!

PR #83, opened 2018-12-06
Send the status bar height for all events

jkmassel:
We need the data on how many users https://github.com/wordpress-mobile/WordPress-iOS/issues/823) is affecting – from there we’ll make the decision whether to mark as `wontfix`, or to redesign the launch screen.


jkmassel:
Thanks @astralbodies !

PR #87, opened 2019-01-04
Add macOS support back to Tracks

jkmassel:
Re-adds macOS support back to Tracks, along with a demo app for testing (similar to what exists for iOS – it allows us to ensure the implementation doesn't break anything on macOS when we're doing iOS-related development).

It bumps the requirement for Tracks to macOS 10.11, which is the same as the minimum version required for Simplenote.


jkmassel:
Thanks for the notes!!

The pod warning message was fixed in https://github.com/Automattic/Automattic-Tracks-iOS/pull/87/commits/ad83cc480dbdd9fa7f2c289aed55ae83461ca14 – needed to move the platform target forward. This also resolved the checksum issue!

The warning about framework targets was resolved by adding the test suite to the macOS framework – which was probably a good thing to do anyways! (see: https://github.com/Automattic/Automattic-Tracks-iOS/pull/87/commits/de620c6d4450698946b9915d595722331886e30e)

Headers are removed – rookie move by me!!

LMK if this looks good from your side, and thanks! 😃

PR #90, opened 2019-01-28
Adding device orientation property

kwonye:
We want to know what orientation the user is currently using when sending over tracks events. This adds it to our device information.

@jkmassel it looks like we use tags for versioning this? If so, can I simply make it 0.3.1 when this gets approved?


kwonye:
Thanks @jkmassel!

PR #93, opened 2019-02-13
Add is_online custom event property

shiki:
This is the first part of wordpress-mobile/WordPress-iOS#11003. 

The next step is updating the WordPress-iOS `Podfile` to use the version of this library with this change.

## Testing

This can be tested by using this library directly in WordPress-iOS.

1. Set this up as a development pod in WordPress-iOS. Replace the current `Automattic-Tracks-iOS` declaration. See [example](https://guides.cocoapods.org/using/the-podfile.html#using-the-files-from-a-folder-local-to-the-machine).
2. Run the WordPress-iOS app. Cleaning the build beforehand would be a good idea. I noticed that, sometimes, Xcode would still use the old pod version (weird).
3. Play around with the app while offline so events are logged. I'm not sure if Reachability works well in a Simulator so testing on a device is probably better.
4. Check the logged events of your WordPress account in Tracks Live View. Pick one event and validate that it has a `device_info_is_online` property. Please note that it takes a while for Tracks to display the logged events. 

We probably won't be able to test this with Tracks Trends until the next day (p4qSXL-2Yz-p2) so the Tracks Live View is what we can test for now.

### DIFF
--- a/Automattic-Tracks-iOS/TracksDeviceInformation.h
+++ b/Automattic-Tracks-iOS/TracksDeviceInformation.h
@@ -18,6 +18,10 @@
 @property (nonatomic, readonly) NSString *currentNetworkOperator;
 @property (nonatomic, readonly) NSString *currentNetworkRadioType;
 @property (nonatomic, assign) BOOL isWiFiConnected;
+/**
+ * Indicates whether the device has an Internet connection.
+ */
+@property (nonatomic, assign) BOOL isOnline;
 @property (nonatomic, assign) BOOL isAppleWatchConnected;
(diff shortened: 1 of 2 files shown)


yaelirub:
@shiki , at this point is there anything else to test other than the project builds successfully?

shiki:
@yaelirub My apologies. I have updated the description to include a testing guide. Please let me know if you have any questions. 🙂

yaelirub:
Followed the testing instructions and seems to work as expected:
screen shot 2019-02-14 at 11 04 53 am

PR #100, opened 2019-04-08
Fix incorrect device_info_is_online values on first few events

shiki:
This is part of wordpress-mobile/WordPress-iOS#11395.

This fixes an issue where we're getting too many `NO` results for the `device_info_is_online` property. This tends to happen for events like `my_site_tab_accessed`, `login_accessed`, and `notifications_notification_details_opened`.

See paCBwp-54-p2 for more info. 

## Findings

After some investigation, I found that on `TracksService`:

1. Every time we send the events to the server, the method `mutableDeviceProperties` is called to acquire the value for custom props like `device_info_is_online` (`deviceInformation.isOnline`).
2. The `deviceInformation.isOnline` is only initialized in `-reachabilityChanged:(NSNotification *)notification`. It is `NO` by default.

The problem is, some events like `my_site_tab_accessed` are sent **before** the `isOnline` property is ever initialized. This causes those events to incorrectly report `device_info_is_online` as `NO`.

## Solution

This PR updates the `isOnline` (and additionally `isWifiConnected`) properties when the `TracksService` is initialized. When I tested it, it looks like `Reachability` reports the correct `isReachable` value during this time.

## Testing

Please test on a device if possible.

1. Check out the branch of this PR: wordpress-mobile/WordPress-iOS#11417
2. Run the WPiOS app while online. Cleaning the build beforehand would be a good idea.
3. Check the logged `my_site_tab_accessed` event in Tracks Live View. Make sure its `device_info_is_online` property is `YES`.
4. Repeat Step 2 and 3 but this time, run the app while **offline**. The `device_info_is_online` should be `NO`.

Please do some regression test the other events as well.

## Others

I believe the risk of regression is low for this so I will target WP iOS 12.2 in wordpress-mobile/WordPress-iOS#11417 when this PR is merged.

### DIFF
--- a/Automattic-Tracks-iOS/TracksService.m
+++ b/Automattic-Tracks-iOS/TracksService.m
@@ -78,8 +78,12 @@ NSString *const USER_ID_ANON = @"anonId";
         _contextManager = contextManager;
         _tracksEventService = [[TracksEventService alloc] initWithContextManager:contextManager];
         _deviceInformation = [TracksDeviceInformation new];
+
         _reachability = [Reachability reachabilityWithHostname:@"public-api.wordpress.com"];
         [_reachability startNotifier];
+
+        [self updateDeviceInformationFromReachability];
+
         _isHostReachable = YES;
         _timerEnabled = YES;
         _userProperties = [NSMutableDictionary new];
@@ -221,6 +225,17 @@ NSString *const USER_ID_ANON = @"anonId";
     }
 }
 
+#pragma mark - Reachability
+
+/**
+ * Update `self.deviceInformation` properties that rely on `self.reachability`.
+ */
+- (void)updateDeviceInformationFromReachability
+{
+    self.deviceInformation.isWiFiConnected = self.reachability.isReachableViaWiFi;
+    self.deviceInformation.isOnline = self.reachability.isReachable;
+}
+
 #pragma mark - Private methods
 
 - (void)didEnterBackground:(NSNotification *)notification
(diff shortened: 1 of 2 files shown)


Automattic-Tracks-iOS/TracksService.m
```diff
@@ -264,6 +260,22 @@ - (void)reachabilityChanged:(NSNotification *)notification
     }
 }
 
+#pragma mark - Private methods
```
shiki: This block is just a re-org of the methods so the Reachability-related methods are grouped together. No changes in content. See commit 2ae92a152759ef2f1c45068f38e933779fede720.

Automattic-Tracks-iOS/TracksService.m
```diff
@@ -78,8 +78,12 @@ - (instancetype)initWithContextManager:(TracksContextManager *)contextManager
         _contextManager = contextManager;
         _tracksEventService = [[TracksEventService alloc] initWithContextManager:contextManager];
         _deviceInformation = [TracksDeviceInformation new];
+
         _reachability = [Reachability reachabilityWithHostname:@"public-api.wordpress.com"];
         [_reachability startNotifier];
+
+        [self updateDeviceInformationFromReachability];
```
shiki: This is the fix. Done in commit 213cd2cf4587f1f200d4b452e909ab1a5fbb4384.

shiki:
Thanks, @jklausa!

PR #106, opened 2019-08-29
Don’t track memory pressure as an event in Sentry

jkmassel:
Pretty simple!

**To Test**
1.  Give the code a read. It's a pretty straightforward change :) 
2. Ensure that all tests are green.


jleandroperez:
Thank you for doing this @jkmassel!!

PR #114, opened 2019-10-25
Disable automatic breadcrumb tracking by default

jkmassel:
After this change, apps must explicitly opt in to breadcrumb tracking – if they don't, nothing happens.

This PR can be tested with https://github.com/wordpress-mobile/WordPress-iOS/pull/12778


frosty (approved):
Just one clarifying question, which I added to the WPiOS PR 🙂

PR #116, opened 2019-12-11
Fix memory layout issues with DDLogLevel

jkmassel:
Rather than using `int` as the type, we use `DDLogLevel` per https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Documentation/DynamicLogLevels.md.

We'll need to merge https://github.com/Automattic/Automattic-Tracks-iOS/pull/117 and rebase this branch before CI will pass for this PR.


jkmassel:
Tests are now passing following the rebase, and given the 👍 on the code from @koke, I'm confident it's ok to merge this.

PR #118, opened 2020-02-18
Fix Core Data warnings.

etoledom:
The warning reads as:

>[error] fault: One or more models in this application are using transformable properties with transformer names that are either unset, or set to NSKeyedUnarchiveFromDataTransformerName. Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead. At some point, Core Data will default to using "NSSecureUnarchiveFromData" when nil is specified, and transformable properties containing classes that do not support NSSecureCoding will become unreadable.

I have followed the proposed solution, using `NSSecureUnarchiveFromData` as the transformer name.

To test:
- CI Tests should pass.
- Check out https://github.com/wordpress-mobile/WordPress-iOS/pull/13489 and follow the test steps there.

### DIFF
--- a/Automattic-Tracks-iOS/Tracks.xcdatamodeld/Tracks.xcdatamodel/contents
+++ b/Automattic-Tracks-iOS/Tracks.xcdatamodeld/Tracks.xcdatamodel/contents
@@ -1,19 +1,19 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<model userDefinedModelVersionIdentifier="" type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="6751" systemVersion="14C1514" minimumToolsVersion="Xcode 4.3" macOSVersion="Automatic" iOSVersion="Automatic">
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="15702" systemVersion="19D76" minimumToolsVersion="Xcode 4.3" sourceLanguage="Objective-C" userDefinedModelVersionIdentifier="">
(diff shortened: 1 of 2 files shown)


etoledom:
@astralbodies - could I bug you with a quick look at this? Thanks!

jkmassel:
This is super good – thanks for tackling it @etoledom!! 

One test case I'd like to propose – does this all work correctly if you:

1) Create a tracks event with the code currently in `develop`.
2) Switch to this branch
3) Try to read and transmit it? 

I'd looked at this issue before and couldn't find any documentation as to whether `NSSecureUnarchiveFromData` is backward-compatible with existing data. If it's not, I worry it might cause a crash-on-startup bug either when it tries to initialize the Core Data store or when it attempts to read the record into the `TracksEvent` object. 

It's hopefully nothing, but this had been in the back of my head for a while, so now that you've fixed it, just wanted to ensure this wouldn't be an issue :)

etoledom:
Thank you @jkmassel for the extra test case. This is what I have tried:

- Delete the app and build from `develop`.
- Wait for the events flush.
- Trigger a new event (in this case switching gutenberg on/off, just because :] )
- Kill the app from Xcode to avoid flushing the events.
- Checkout and build from this branch.
- On app launch, wait for the first events flush.
- Check that the gutenberg switch event is there, and it's sent to the backend.
- Success 🎉 

After performing this test, everything seems to behave as expected 👍

jkmassel:
Thanks for indulging my worry @etoledom!!

astralbodies:
@etoledom I really apologize for the delay on this review, I will have it to you by tomorrow!

PR #123, opened 2020-03-18
Fixes code signing during development

jkmassel:
Depends on #122 – it should be rebase-merged before this PR is reviewed.

**This PR:**
- Removes all references to Development Teams, Code Signing Identities, and disables automatic code signing
- Disables Catalyst support (until it's available in Sodium – we don't need it right now though, and enabling it breaks tests)

**To Test:**
- Ensure CI Passes
- Ensure the framework builds and compiles properly locally and that you can run iOS and macOS tests locally without signing errors


jkmassel:
@shiki – this is ready for review now!

PR #124, opened 2020-03-18
Reorganize the project

jkmassel:
Organizes the project file system to match the project file before we make too many more changes. I've broken this down into several smaller commits to group things changes by folder in order to make review easier.

**To Test**
- Ensure CI passes
- Test locally if you'd like

**Future Steps**
The majority of code doesn't _actually_ belong to the iOS target, but it's there because that one came first. It might make sense to extract the majority of the code to a `shared` (and `sharedTests`?) directory. WDYT?


jkmassel:
>I don't think these belong inside the Model folder. TracksContextManager is an important class that should be easily discoverable especially because it's used directly by consumers. Perhaps a top-level Core Data folder would be better.

Agreed – perhaps that could be a future change.

>My gut feeling is that there should be a shared "project", not just a directory. But I don't have a lot of experience with how that works so I can't really say if that will work well.

That would probably be a good way to go – CocoaPods doesn't really care about the Xcode project – just the files, so we can organize it sort of however we want.

PR #127, opened 2020-03-23
Add event logging upload

jkmassel:
Adds support for uploading encrypted logs. 

Assigned to both @oguzkocer and @shiki, though either (or both) can approve and provide comments as desired.

In the scope of review, I'd love if you're able to comment specifically on whether:
- All public method documentation is easy to read and understand.
- The use of preconditions + Implicitly Unwrapped Optionals in `EventLoggingUploadManager` seems reasonable. Not providing a delegate and a dataSource is a programmer error, so I used these to avoid a bunch of runtime error handling code that should never be called anyway.

**To Test:**
- Ensure CI passes
- Ensure tests pass locally

### DIFF
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
@@ -1,0 +1,33 @@
+import Foundation
+
+public protocol EventLoggingDelegate {
+    /// The event logging system will call this delegate property prior to attempting to upload, giving the application a chance to determine
+    /// whether or not the upload should proceed. If this is not overridden, the default is `false`.
+    var shouldUploadLogFiles: Bool { get }
+
+    /// The event logging system will call this delegate method each time a log file starts uploading.
+    func didStartUploadingLog(_ log: LogFile)
+
+    /// The event logging system will call this delegate method if a log file upload is cancelled by the delegate.
+    func uploadCancelledByDelegate(_ log: LogFile)
+
+    /// The event logging system will call this delegate method if a log file fails to upload.
+    func uploadFailed(withError: Error, forLog: LogFile)
+
+    /// The event logging system will call this delegate method each time a log file finishes uploading.
+    func didFinishUploadingLog(_ log: LogFile)
+}
+
+/// Default implementations for EventLoggingDelegate
+public extension EventLoggingDelegate {
+
+    var shouldUploadLogFiles: Bool {
+        return false // Use a privacy-preserving default
+    }
+
+    // Empty default implementations allow the developer to only implement these if they need them
+    func didStartUploadingLog(_ log: LogFile) {}
+    func uploadCancelledByDelegate(_ log: LogFile) {}
+    func uploadFailed(withError error: Error, forLog log: LogFile) {}
+    func didFinishUploadingLog(_ log: LogFile) {}
+}
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
@@ -12,9 +12,11 @@ public protocol EventLoggingDelegate {
     func uploadCancelledByDelegate(_ log: LogFile)
 
     /// The event logging system will call this delegate method if a log file fails to upload.
+    /// It may be called prior to upload starting if the file is missing, and is called prior to the `upload` callback.
     func uploadFailed(withError: Error, forLog: LogFile)
 
     /// The event logging system will call this delegate method each time a log file finishes uploading.
+    /// It is called prior to the `upload` callback.
     func didFinishUploadingLog(_ log: LogFile)
 }
 
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
@@ -4,7 +4,7 @@ import Sodium
 
 class EventLoggingUploadManager {
 
-    private struct Constants {
+    private enum Constants {
         static let uuidHeaderKey = "log-uuid"
         static let uploadHttpMethod = "POST"
     }
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
@@ -27,7 +27,7 @@ class EventLoggingUploadManager {
         }
 
         guard let fileContents = FileManager.default.contents(atUrl: log.url) else {
-            delegate.uploadFailed(withError: UploadError.fileMissing, forLog: log)
+            delegate.uploadFailed(withError: EventLoggingFileUploadError.fileMissing, forLog: log)
             return
         }
 
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
@@ -9,24 +9,25 @@ class EventLoggingUploadManager {
         static let uploadHttpMethod = "POST"
     }
 
-    internal var networkService = EventLoggingNetworkService()
-
-    internal var dataSource: EventLoggingDataSource!
-    internal var delegate: EventLoggingDelegate!
+    var dataSource: EventLoggingDataSource
+    var delegate: EventLoggingDelegate
+    var networkService: EventLoggingNetworkService = EventLoggingNetworkService()
+    var fileManager: FileManager = FileManager.default
 
     typealias LogUploadCallback = (Result<Void, Error>) -> Void
 
+    init(dataSource: EventLoggingDataSource, delegate: EventLoggingDelegate) {
+        self.dataSource = dataSource
+        self.delegate = delegate
+    }
+
     func upload(_ log: LogFile, then callback: @escaping LogUploadCallback) {
-
-        precondition(delegate != nil , "You must set the Event Logging Delegate prior to attempting an upload")
-        precondition(dataSource != nil , "You must set the Event Logging Data Source prior to attempting an upload")
-
         guard delegate.shouldUploadLogFiles else {
-            delegate?.uploadCancelledByDelegate(log)
+            delegate.uploadCancelledByDelegate(log)
             return
         }
 
-        guard let fileContents = FileManager.default.contents(atUrl: log.url) else {
+        guard let fileContents = fileManager.contents(atUrl: log.url) else {
             delegate.uploadFailed(withError: EventLoggingFileUploadError.fileMissing, forLog: log)
             return
         }
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
@@ -46,7 +46,7 @@ class EventLoggingUploadManager {
         delegate.didStartUploadingLog(log)
 
         networkService.uploadFile(request: request, fileURL: log.url) { result in
-            switch(result) {
+            switch result {
                 case .success:
                     self.delegate.didFinishUploadingLog(log)
                     callback(.success(()))
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
@@ -1,0 +1,52 @@
+import Foundation
+import CocoaLumberjack
+import Sodium
+
+class EventLoggingUploadManager {
+
+    private struct Constants {
+        static let uuidHeaderKey = "log-uuid"
+        static let uploadHttpMethod = "POST"
+    }
+
+    internal var networkService = EventLoggingNetworkService()
+
+    internal var dataSource: EventLoggingDataSource!
+    internal var delegate: EventLoggingDelegate!
+
+    typealias LogUploadCallback = (Result<Void, Error>) -> Void
+
+    func upload(_ log: LogFile, then callback: @escaping LogUploadCallback) {
+
+        precondition(delegate != nil , "You must set the Event Logging Delegate prior to attempting an upload")
+        precondition(dataSource != nil , "You must set the Event Logging Data Source prior to attempting an upload")
+
+        guard delegate.shouldUploadLogFiles else {
+            delegate?.uploadCancelledByDelegate(log)
+            return
+        }
+
+        guard let fileContents = FileManager.default.contents(atUrl: log.url) else {
+            delegate.uploadFailed(withError: UploadError.fileMissing, forLog: log)
+            return
+        }
+
+        var request = URLRequest(url: dataSource.logUploadURL)
+        request.addValue(log.uuid, forHTTPHeaderField: Constants.uuidHeaderKey)
+        request.httpMethod = Constants.uploadHttpMethod
+        request.httpBody = fileContents
+
+        delegate.didStartUploadingLog(log)
+
+        networkService.uploadFile(request: request, fileURL: log.url) { result in
+            switch(result) {
+                case .success:
+                    self.delegate.didFinishUploadingLog(log)
+                    callback(.success(()))
+                case .failure(let error):
+                    self.delegate.uploadFailed(withError: error, forLog: log)
+                    callback(.failure(error))
+            }
+        }
+    }
+}
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
@@ -9,16 +9,22 @@ class EventLoggingUploadManager {
         static let uploadHttpMethod = "POST"
     }
 
-    var dataSource: EventLoggingDataSource
-    var delegate: EventLoggingDelegate
-    var networkService: EventLoggingNetworkService = EventLoggingNetworkService()
-    var fileManager: FileManager = FileManager.default
+    private let dataSource: EventLoggingDataSource
+    private let delegate: EventLoggingDelegate
+    private let networkService: EventLoggingNetworkService
+    private let fileManager: FileManager
 
     typealias LogUploadCallback = (Result<Void, Error>) -> Void
 
-    init(dataSource: EventLoggingDataSource, delegate: EventLoggingDelegate) {
+    init(dataSource: EventLoggingDataSource,
+         delegate: EventLoggingDelegate,
+         networkService: EventLoggingNetworkService = EventLoggingNetworkService(),
+         fileManager: FileManager = FileManager.default
+    ) {
         self.dataSource = dataSource
         self.delegate = delegate
+        self.networkService = networkService
+        self.fileManager = fileManager
     }
 
     func upload(_ log: LogFile, then callback: @escaping LogUploadCallback) {
--- a/Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
+++ b/Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
@@ -1,0 +1,27 @@
+import Foundation
+
+extension FileManager {
+
+    var documentsDirectory: URL {
+        let documentsDirectory = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first!
+        return URL(fileURLWithPath: documentsDirectory, isDirectory: true)
+    }
+
+    func contentsOfDirectory(at url: URL) throws -> [URL] {
+        return try FileManager.default.contentsOfDirectory(atPath: url.path).map { url.appendingPathComponent($0) }
+    }
+
+    func contents(atUrl url: URL) -> Data? {
+        return self.contents(atPath: url.path)
+    }
+
+    func fileExistsAtURL(_ url: URL) -> Bool {
+        return FileManager.default.fileExists(atPath: url.path)
+    }
+
+    func directoryExistsAtURL(_ url: URL) -> Bool {
+        var isDir: ObjCBool = false
+        let exists = FileManager.default.fileExists(atPath: url.path, isDirectory: &isDir)
+        return exists && isDir.boolValue
+    }
+}
--- a/Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
+++ b/Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
@@ -8,7 +8,7 @@ extension FileManager {
     }
 
     func contentsOfDirectory(at url: URL) throws -> [URL] {
-        return try FileManager.default.contentsOfDirectory(atPath: url.path).map { url.appendingPathComponent($0) }
+        return try self.contentsOfDirectory(atPath: url.path).map { url.appendingPathComponent($0) }
     }
 
     func contents(atUrl url: URL) -> Data? {
@@ -16,12 +16,12 @@ extension FileManager {
     }
 
     func fileExistsAtURL(_ url: URL) -> Bool {
-        return FileManager.default.fileExists(atPath: url.path)
+        return self.fileExists(atPath: url.path)
     }
 
     func directoryExistsAtURL(_ url: URL) -> Bool {
         var isDir: ObjCBool = false
-        let exists = FileManager.default.fileExists(atPath: url.path, isDirectory: &isDir)
+        let exists = self.fileExists(atPath: url.path, isDirectory: &isDir)
         return exists && isDir.boolValue
     }
 }
--- a/Automattic-Tracks-iOS/Model/UploadError.swift
+++ b/Automattic-Tracks-iOS/Model/UploadError.swift
@@ -1,0 +1,13 @@
+import Foundation
+
+public enum UploadError: Error, LocalizedError {
+    case httpError(String)
+    case fileMissing
+
+    public var errorDescription: String?{
+        switch self {
+            case .httpError(let message): return message
+            case .fileMissing: return "File not found"
+        }
+    }
+}
--- a/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
@@ -1,0 +1,32 @@
+import Foundation
+
+typealias NetworkResultCallback = (Result<Data?, Error>) -> Void
+
+open class EventLoggingNetworkService {
+
+    func uploadFile(request: URLRequest, fileURL: URL, completion: @escaping NetworkResultCallback) {
+        URLSession.shared
+            .uploadTask(with: request, fromFile: fileURL, completionHandler: { data, response, error in
+
+            if let error = error {
+                completion(.failure(error))
+                return
+            }
+
+            /// The `response` should *always* be an HTTPURLResponse, but we'll be defensive anyway
+            guard let statusCode = (response as? HTTPURLResponse)?.statusCode else {
+                completion(.success(data))
+                return
+            }
+
+            /// Generate a reasonable error message based on the HTTP status
+            if !(200 ... 299).contains(statusCode) {
+                let errorMessage = HTTPURLResponse.localizedString(forStatusCode: statusCode)
+                completion(.failure(UploadError.httpError(errorMessage)))
+                return
+            }
+
+            completion(.success(data))
+        }).resume()
+    }
+}
--- a/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
@@ -4,9 +4,15 @@ typealias NetworkResultCallback = (Result<Data?, Error>) -> Void
 
 open class EventLoggingNetworkService {
 
+
+    private let urlSession: URLSession
+
+    init(urlSession: URLSession = URLSession.shared) {
+        self.urlSession = urlSession
+    }
+
     func uploadFile(request: URLRequest, fileURL: URL, completion: @escaping NetworkResultCallback) {
-        URLSession.shared
-            .uploadTask(with: request, fromFile: fileURL, completionHandler: { data, response, error in
+        urlSession.uploadTask(with: request, fromFile: fileURL, completionHandler: { data, response, error in
 
             if let error = error {
                 completion(.failure(error))
--- a/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
@@ -13,11 +13,8 @@ open class EventLoggingNetworkService {
                 return
             }
 
-            /// The `response` should *always* be an HTTPURLResponse, but we'll be defensive anyway
-            guard let statusCode = (response as? HTTPURLResponse)?.statusCode else {
-                completion(.success(data))
-                return
-            }
+            /// The `response` should *always* be an HTTPURLResponse. Crash if notl
+            let statusCode = (response as! HTTPURLResponse).statusCode
 
             /// Generate a reasonable error message based on the HTTP status
             if !(200 ... 299).contains(statusCode) {
--- a/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
@@ -1,9 +1,8 @@
 import Foundation
 
-typealias NetworkResultCallback = (Result<Data?, Error>) -> Void
-
 open class EventLoggingNetworkService {
 
+    typealias ResultCallback = (Result<Data?, Error>) -> Void
 
     private let urlSession: URLSession
 
@@ -11,7 +10,7 @@ open class EventLoggingNetworkService {
         self.urlSession = urlSession
     }
 
-    func uploadFile(request: URLRequest, fileURL: URL, completion: @escaping NetworkResultCallback) {
+    func uploadFile(request: URLRequest, fileURL: URL, completion: @escaping ResultCallback) {
         urlSession.uploadTask(with: request, fromFile: fileURL, completionHandler: { data, response, error in
 
             if let error = error {
--- a/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-open class EventLoggingNetworkService {
+class EventLoggingNetworkService {
 
     typealias ResultCallback = (Result<Data?, Error>) -> Void
 
--- a/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
@@ -18,7 +18,8 @@ class EventLoggingNetworkService {
                 return
             }
 
-            /// The `response` should *always* be an HTTPURLResponse. Crash if notl
+            /// The `response` should *always* be an HTTPURLResponse.
+            /// Fail fast by force-unwrapping – this will cause a crash and will bring the issue to our attention if something has changed.
             let statusCode = (response as! HTTPURLResponse).statusCode
 
             /// Generate a reasonable error message based on the HTTP status
--- a/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
@@ -24,7 +24,7 @@ open class EventLoggingNetworkService {
             /// Generate a reasonable error message based on the HTTP status
             if !(200 ... 299).contains(statusCode) {
                 let errorMessage = HTTPURLResponse.localizedString(forStatusCode: statusCode)
-                completion(.failure(UploadError.httpError(errorMessage)))
+                completion(.failure(EventLoggingFileUploadError.httpError(errorMessage)))
                 return
             }
 
--- a/Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-open class EventLoggingUploadQueue {
+class EventLoggingUploadQueue {
 
     private let fileManager: FileManager
     var storageDirectory: URL
--- a/Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
+++ b/Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
@@ -21,14 +21,14 @@ open class EventLoggingUploadQueue {
     }
 
     func add(_ log: LogFile) throws {
-        try ensureStorageDirectoryExists()
+        try createStorageDirectoryIfNeeded()
         try fileManager.copyItem(at: log.url, to: storageURL(forLog: log))
     }
 
     func remove(_ log: LogFile) throws {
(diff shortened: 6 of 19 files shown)


Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
```diff
@@ -0,0 +1,27 @@
...
+
+extension FileManager {
+
+    var documentsDirectory: URL {
+        let documentsDirectory = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first!
+        return URL(fileURLWithPath: documentsDirectory, isDirectory: true)
+    }
+
+    func contentsOfDirectory(at url: URL) throws -> [URL] {
+        return try FileManager.default.contentsOfDirectory(atPath: url.path).map { url.appendingPathComponent($0) }
```
oguzkocer: Do we really want to have the `FileManager.default.` bit in these methods? Shouldn't we call the methods on this particular `FileManager` instead or if we really want the `FileManager.default` (which I don't think we do) shouldn't this be a static function?

This comment applies to other instances of `FileManager.default` in this extension.

Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    func contents(atUrl url: URL) -> Data? {
+        return self.contents(atPath: url.path)
+    }
+
+    func fileExistsAtURL(_ url: URL) -> Bool {
+        return FileManager.default.fileExists(atPath: url.path)
+    }
```
oguzkocer: I realize this is a very common thing in iOS land, but I am still failing to see the value in having these extensions instead of directly calling them in the code. Isn't this adding more cognitive load to whoever is looking at this code because they now also need to look into how it's implemented?

Anyway, this is not really a request to change or even a nitpicky comment, just curious about why this is the common approach.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+
+open class EventLoggingUploadQueue {
```
oguzkocer: Similar to my previous comment, does this need to be an `open class`?

Also, this class doesn't seem to be used, I am guessing you're planning to use it in a different PR.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    private let fileManager = FileManager.default
```
oguzkocer: I think it'd be better to take this as a constructor parameter instead so the caller can customize it to their needs. It'd also make this class more testable.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    func remove(_ log: LogFile) throws {
+        let url = storageURL(forLog: log)
+        if fileManager.fileExistsAtURL(url) {
+            try fileManager.removeItem(at: storageURL(forLog: log))
+        }
+    }
+
+    var storageDirectory: URL {
```
oguzkocer: Question: Is there a convention to keep the properties at the top of the file in iOS? That's what we do in Android, but it might be different here.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
...
+        return fileManager.documentsDirectory.appendingPathComponent("log-upload-queue")
```
oguzkocer: How do you feel about making the `log-upload-queue` a constructor parameter instead to make this class more customizable? If we do that, we can probably make the a stored property instead as well.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    func storageURL(forLog log: LogFile) -> URL {
+        let newURL = storageDirectory.appendingPathComponent(log.uuid)
+        return newURL
```
oguzkocer: Nitpick: We don't need to local variable here.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    func ensureStorageDirectoryExists() throws {
```
oguzkocer: Nitpick: Before checking the implementation of this method I thought that it was going to crash if the directory didn't exist. How do you feel about renaming it to something like `createStorageDirectoryIfItDoesntExist` instead?

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    func add(_ log: LogFile) throws {
+        try ensureStorageDirectoryExists()
+        try fileManager.copyItem(at: log.url, to: storageURL(forLog: log))
+    }
+
+    func remove(_ log: LogFile) throws {
+        let url = storageURL(forLog: log)
+        if fileManager.fileExistsAtURL(url) {
+            try fileManager.removeItem(at: storageURL(forLog: log))
```
oguzkocer: Can we not use the local variable `url` here instead?

```suggestion
            try fileManager.removeItem(at: url)
```

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
```diff
@@ -0,0 +1,43 @@
...
+    private let fileManager = FileManager.default
+
+    /// The log file on top of the queue
+    var first: LogFile? {
+
+        guard let url = try? fileManager.contentsOfDirectory(at: storageDirectory).first else {
+            return nil
+        }
+
+        return LogFile(url: url, uuid: url.lastPathComponent)
```
oguzkocer: I am sure this would be fine for a lot of people who'd review this PR, but the thing is, this little bit of code assumes knowledge of how something is implemented in another part of the library. In these type of cases, I always suggest refactoring the code a bit to remove the extra cognitive load.

In this case, the storage path creation happens in this class, so I'd be totally fine to leave it as is. However, if you think it'd be a good improvement, I'd suggest maybe an extension method to `LogFile` in this file with 2 methods, one creating the `storageUrl` and the other one creating a logfile from a given storage url, so the pairing is in the same place.

Again, I am just sharing how I think about these things and a possible improvement that might make an impact in a lot of cases but has a very limited one in this case. Hope that makes sense!

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import CocoaLumberjack
+import Sodium
+
+class EventLoggingUploadManager {
+
+    private struct Constants {
```
oguzkocer: I've been looking into this and it looks like this creating `struct` method to hold onto constants is more common for sharing things between files. So, I was curious why you opted for this in this case? I thought maybe it was to keep relevant constants together but then `Constants` name would not fit.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -0,0 +1,52 @@
...
+        static let uuidHeaderKey = "log-uuid"
+        static let uploadHttpMethod = "POST"
+    }
+
+    internal var networkService = EventLoggingNetworkService()
+
+    internal var dataSource: EventLoggingDataSource!
+    internal var delegate: EventLoggingDelegate!
```
oguzkocer: I am not sure why these are `internal`, but how would you feel about getting them as constructor parameters instead? Similar to my previous comments (in terms of when I wrote them, but I know this will probably show up first) this would make it easier to test this class and also make it more customizable. We could even make these non-nullable this way (if we want) and get rid of the `precondition`s. What do you think?

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -0,0 +1,52 @@
...
+
+        precondition(delegate != nil , "You must set the Event Logging Delegate prior to attempting an upload")
+        precondition(dataSource != nil , "You must set the Event Logging Data Source prior to attempting an upload")
+
+        guard delegate.shouldUploadLogFiles else {
+            delegate?.uploadCancelledByDelegate(log)
+            return
+        }
+
+        guard let fileContents = FileManager.default.contents(atUrl: log.url) else {
```
oguzkocer: I'd personally take `FileManager` as a constructor parameter as well. However, I understand that it's very common to use the `FileManager.default`.

Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
```diff
@@ -0,0 +1,33 @@
...
+    func didStartUploadingLog(_ log: LogFile)
+
+    /// The event logging system will call this delegate method if a log file upload is cancelled by the delegate.
+    func uploadCancelledByDelegate(_ log: LogFile)
+
+    /// The event logging system will call this delegate method if a log file fails to upload.
+    func uploadFailed(withError: Error, forLog: LogFile)
+
+    /// The event logging system will call this delegate method each time a log file finishes uploading.
+    func didFinishUploadingLog(_ log: LogFile)
```
oguzkocer: We may want to add a comment explaining that these delegates will be called before the actual callback goes through in `EventLoggingUploadManager`. I am not too familiar with the naming and conventions in iOS (anymore) so I'd make sure that the ordering of these callbacks and their names match the SDK to avoid confusion.

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+typealias NetworkResultCallback = (Result<Data?, Error>) -> Void
+
+open class EventLoggingNetworkService {
```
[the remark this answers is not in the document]
jkmassel: We could use a `protocol`, but when it's time to use this, we'd need a concrete implementation. It'd be possible to confirm an object to this protocol and thus adopt this method as a default implementation, but I'm not sure I'm a fan of that.

The `open` part just allows other packages (like the test package) to subclass this – classes in Swift are `final` by default outside their package, which is why this one is open.

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
```diff
@@ -0,0 +1,32 @@
...
+            .uploadTask(with: request, fromFile: fileURL, completionHandler: { data, response, error in
+
+            if let error = error {
+                completion(.failure(error))
+                return
+            }
+
+            /// The `response` should *always* be an HTTPURLResponse, but we'll be defensive anyway
+            guard let statusCode = (response as? HTTPURLResponse)?.statusCode else {
+                completion(.success(data))
```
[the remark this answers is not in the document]
jkmassel: The reason for this is that `error` will be populated if there was an upload error, so if it's not, it seems safe to assume that the response was successful.

I also prefer to crash in these situations, so I'm very happy to refactor this. 

Changed in 4f20dc627722d58a87d15604103cd16978541198

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+typealias NetworkResultCallback = (Result<Data?, Error>) -> Void
+
+open class EventLoggingNetworkService {
+
+    func uploadFile(request: URLRequest, fileURL: URL, completion: @escaping NetworkResultCallback) {
+        URLSession.shared
```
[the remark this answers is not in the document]
jkmassel: Yep – this makes sense. The mocks work outside of `URLSession`, but I could see a case where a user of this framework wants to implement a `URLSessionConfiguration` from the parent app on this library, so this is great.

Changed in fac0f65f1637d9add3779a77b5ee3a22a654a3ab

Automattic-Tracks-iOS/Model/UploadError.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public enum UploadError: Error, LocalizedError {
```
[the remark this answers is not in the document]
jkmassel: Yep, no objection – I also thought about moving it inside EventLoggingNetworkService since it's only emitted from there, but I'm not sure if we have a style preference for this?

I went with an even-more-specific `EventLoggingFileUploadError`.

Changed in 6b135c5a6280eae2dc9a30c5be904cfc2d214e94

Automattic-Tracks-iOS/Model/UploadError.swift
```diff
@@ -0,0 +1,13 @@
...
+    case httpError(String)
+    case fileMissing
+
+    public var errorDescription: String?{
```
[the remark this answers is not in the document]
jkmassel: In this case it does, yep, because the object conforms to `LocalizedError`, [which specifies all of its strings as optional](https://developer.apple.com/documentation/foundation/localizederror?changes=_4). 

And nice catch – we should get SwiftLint going on this repo.

Automattic-Tracks-iOS/Model/UploadError.swift
```diff
@@ -0,0 +1,13 @@
...
+        switch self {
+            case .httpError(let message): return message
+            case .fileMissing: return "File not found"
```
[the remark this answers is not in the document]
jkmassel: Hmmm – good point. It _shouldn't_ need it, but we can't guarantee that, so I've enabled translation for it just in case in ac7532802aef539c248490c9f10a142817194b70.

Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
jkmassel: Ah – you're totally right about this – we should use the same instance.

Fixed in de728bd345be5c5615211fb0b03cbb3bbd6a94fb.

Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
jkmassel: Good question – by default, `FileManager` likes to work with paths, which are just a `String`.  Unfortunately, it's _really_ easy to mess up (NS)URL => String conversion (do I use `absoluteString`? `relativeString`? `path`? `relativePath`?)  – there's a _lot_ of properties on `URL`, so not having to think about these conversions at the call sites, and being able to perform operations on `FileManager` in a type-safe way like this (just always represent paths as URLs and you can never do it wrong) is pretty beneficial, even if it has some silly aliases like this.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: Same justification about it being `open`. 

Yep – the next PR will handle putting things on the queue, then taking them off uploading them.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: This is a good point – rather than allow injecting the title, I think it makes more sense to allow a developer to inject the entire `storageDirectory` – perhaps they'd rather store them in temporary storage rather than in the user documents directory? 

Adjusted in 810b5c9124ad66e6eff3d08c8e714ebf460bf496.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: Snuck into 810b5c9124ad66e6eff3d08c8e714ebf460bf496.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: Good question – for "normal" properties, I agree that this makes sense.

For the case of calculated properties like this, because they could be represented as `func storageDirectory() -> URL`, IMHO it's essentially a method, and so I like not cluttering up the top of my class with it, but I'm not aware of a code style rule for this.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: Good call – adjusted with https://github.com/Automattic/Automattic-Tracks-iOS/commit/810b5c9124ad66e6eff3d08c8e714ebf460bf496.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: Works for me – changed in 92afe2d422968422c12f367bfe0d1b09cce1c3e9.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: Ah yep! Fixed with 92afe2d422968422c12f367bfe0d1b09cce1c3e9

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
jkmassel: Yep, it was just to keep the constants for this class together – I could've just exposed them as `static let ...` inside the class, but this just cut down the noise and made them pretty obvious when reading – basically it's just a way to namespace them out of the way.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
jkmassel: That's a better solution for sure. It'll require re-working the wrapping API for this, but this will force a better solution for ensuring they're present. I probably should've thought of this!! 🙃

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
jkmassel: I reworked this to allow the developer to customize the instance of `FileManager` used, but putting it in the constructor seemed like a lot – it'd be exceedingly rare to use an alternative FileManager (I can't think of a file I've ever done it, and I couldn't find a single instance of it in WPiOS code – with the exception of one third-party Pod), so it would seem like clutter IMHO. 

Done in 4ea3255a864ec7f5b9cd22114162eddcbdb8fed9

Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
jkmassel: Good call on documenting it. I think we might be hard pressed for a definitive answer on which it should be, but my thinking was to fire the delegate method as soon as we know whether the request was successful or not (ie – `didFinish` is called as soon as it finishes).

I'm open to discussion on whether this is the best way, but IMHO the delegate should have the most up-to-date info since it's event-based and it's ok for the callback to lag behind – there's no expectation that'll be called instantly following a result, just that it _will_ be called in the future.

Docs updated in 5e34b4ef2d29f14a3c57d82c51a5e406c72edff1.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
jkmassel: Yeah this was a tricky one! I'm not big on classes having any input into their persistence, so this seemed clean. My thought was that if we needed to persist it elsewhere we could have a `LogFileStore` that wrangles that, but it'd be overkill to have it for just one instance. 

I've reworked it to allow initialization without this implementation detail leaking, as well as adding a corresponding `fileName` property that consumers can use to persist it. This hides the implementation details a bit more nicely. 

WDYT? Implemented in 0f92472a683191817f3847479b0959fdc09e2cbd.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
shiki: You both made a good decision here. 🙂

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    private struct Constants {
+        static let uuidHeaderKey = "log-uuid"
+        static let uploadHttpMethod = "POST"
+    }
+
+    var dataSource: EventLoggingDataSource
+    var delegate: EventLoggingDelegate
+    var networkService: EventLoggingNetworkService = EventLoggingNetworkService()
+    var fileManager: FileManager = FileManager.default
```
shiki: I'd also consider making these `private let`. I find that it's easier to understand how to use a class if it has few public methods. 

```suggestion
    private let dataSource: EventLoggingDataSource
    private let delegate: EventLoggingDelegate
    private let networkService: EventLoggingNetworkService = EventLoggingNetworkService()
    private let fileManager: FileManager = FileManager.default
```

Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
+public protocol EventLoggingDelegate {
```
shiki: The [WP Swift Style Guide](https://github.com/wordpress-mobile/swift-style-guide) says that we try to follow the  [raywenderlich.com Swift Style Guide](https://github.com/raywenderlich/swift-style-guide/) and that style guide mentions this for [naming delegate methods](https://github.com/raywenderlich/swift-style-guide/#delegates): 

> When creating custom delegate methods, an unnamed first parameter should be the delegate source. (UIKit contains numerous examples of this.)

This is also the standard that iOS SDK follows as far as I can tell. 

But that's up to you if you'd like to follow that. I just find that that naming pattern makes it easier to understand where methods come from and how they are used.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
shiki: There's also a 2-year-old proposal by Diego to use an `enum` instead of a `struct` for grouping. https://github.com/wordpress-mobile/swift-style-guide/issues/12

You're welcome for making you more confused. 😄

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+import CocoaLumberjack
+import Sodium
+
+class EventLoggingUploadManager {
```
shiki: Is this the main entry point for the WP/WC apps? If so, should this be `public`?

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
```diff
@@ -0,0 +1,34 @@
+import Foundation
+
+open class EventLoggingNetworkService {
+
+    typealias ResultCallback = (Result<Data?, Error>) -> Void
```
shiki: 👏 for using `Result`

Automattic-Tracks-iOSTests/Test Helpers/Extensions.swift
```diff
@@ -26,3 +27,11 @@ extension FileManager {
         return fileURL
     }
 }
+
+extension XCTestCase {
+    func async_test(timeout: TimeInterval, _ block: (XCTestExpectation) -> ()) {
```
shiki: Not important:

Well done for improving this cumbersome XCTest API.

:twitch: at the underscore 😄 

You could rename this to `waitUntil` instead. Inspired by [Nimble's asynchronous expectations](https://github.com/Quick/Nimble#asynchronous-expectations).

Automattic-Tracks-iOSTests/Tests/Event Logging/EventLoggingNetworkServiceTests.swift
```diff
@@ -0,0 +1,81 @@
...
+        super.setUp()
+        self.service = EventLoggingNetworkService()
+    }
+
+    override func tearDown() {
+        self.service = nil
+        HTTPStubs.removeAllStubs()
+
+        super.tearDown()
+    }
```
shiki: 🤸‍♂

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
jkmassel: Chatting on Slack, we discussed how decreased mutability was preferable to simpler initialization, so these are now all provided via the constructor in a970df2b1c6f215367daf1d784285316a2a90c4d.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
jkmassel: I like the `enum` thing – and it's also suggested in the [RW style guide](https://github.com/raywenderlich/swift-style-guide/#constants).

Changed in 0c5b5bff1ba54eed15524251d91575c8ef89bb18.

Automattic-Tracks-iOSTests/Test Helpers/Extensions.swift
jkmassel: Ugh, totally – I've been writing too much Ruby 😅 

Renamed to `waitForExpectation` – I considered using the same api as Nimble `waitUntil(timeout: TimeInterval, done: () -> Void)`, but this felt more in line with the rest of `XCTest` IMHO, and as this is an extension vs a separate framework I felt like imitating that style made a bit more sense.

Renamed in 5da3755d8c47fe6d9f1bb12494f33552fe96ff78

Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
jkmassel: This is generally a good practice for sure. In this case, I'd like to try avoiding adding `EventLoggingUploadManager` in the callback, because if a developer were to receive it, there are no public methods or properties they could work with – having the object wouldn't really help them.

Additionally, I plan to wrap `EventLoggingUploadManager` and `EventLoggingUploadQueue` in a service object that allows you to initialize all of this without being concerned with the internals – from a developer's perspective, the delegate will appear to belong to that wrapper object.

I think this'll make sense in the context of the final PR, but I'll be sure to leave a note to take another look at this as part of that one.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
jkmassel: Just to reiterate https://github.com/Automattic/Automattic-Tracks-iOS/pull/127#discussion_r398788586, this still isn't part of the public API – that'll come in the next PR :)

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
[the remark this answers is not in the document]
oguzkocer: > The open part just allows other packages (like the test package) to subclass this – classes in Swift are final by default outside their package, which is why this one is open.

Yeap, I know about that part and that is what was bothering me. I don't like implementation inheritance at all and strongly believe there is always a better way to do it. Having said that, I don't try and enforce this subjective opinion and just bring it up and leave it to the developer. If you want to talk about this further I am happy to, but otherwise if you're happy we can totally leave it as is.

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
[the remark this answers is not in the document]
oguzkocer: I was initially fine with this, but then I saw that we are checking the status code and calling the error callback if it's not in `2xx` range. That to me means either `statusCode` is absolutely necessary and we should crash, or we should call error callback if it doesn't exist. Without that bit of code, the original version of how you did this makes more sense to me. Does that make sense?

The change looks good, but there is a typo at the end of the comment. I also try to be more explicit about the comment about why we are crashing if something unexpected happens in these cases just so the next developer understands why we opted to crash. I should have mentioned this in my original comment, but anyway, here is how I would have written this comment:

```
/// The `response` should *always* be an HTTPURLResponse. To follow fail fast principle,
/// we explicitly crash here in case it's something else so we can be aware the issue and
/// address it instead of it going unnoticed.
```

Adding this type of comment in every controlled crash is a bit daunting, but I don't think this is common enough of a pattern in a8c to not do it.

As in most cases, this is just a subjective suggestion, so feel free to keep the comment as is.

Automattic-Tracks-iOS/Model/UploadError.swift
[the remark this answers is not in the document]
oguzkocer: I generally start with a `private enum` in these cases and only move it to a `public enum` if it'll be used elsewhere. I don't think there is any hard & fast rules about this, so whatever makes the most sense for you is fine by me.

Resolving this comment because I don't think any change _needs_ to be made, but feel free to make it, unresolve the comment and add a comment about the change :)

Automattic-Tracks-iOS/Model/UploadError.swift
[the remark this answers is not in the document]
oguzkocer: Oh I see, that's bummer. Thanks for pointing it out!

Automattic-Tracks-iOS/Extensions/FileManager+Extensions.swift
oguzkocer: That makes a lot of sense, thanks a lot for explaining all this to me!

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
oguzkocer: That sounds good to me. I am going to resolve this conversation but there are a couple things in the commit I'd like to point out that I'll do separately.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
oguzkocer: I kind of prefer the original over this 😬 I personally prefer to not have anything mutable unless it absolutely has to be. Like when the SDK forces you to make it mutable type of thing.

It looks like you made some changes to this since the commit, so I am just going to resolve this comment and start a new conversation about this if it's still necessary.

Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
oguzkocer: If there isn't a rule, what I'd think about in this case is "Does the delegate require any information from the execution of the callback?" if so, do the `delegate` after.

For example, let's say that `LogUploadCallback` we pass is marking the state of the log file as `uploaded` and the delegate is reading that state in `didFinishUploadingLog`. In that case, we'd want the `callback` to be called before the `delegate`.

I guess maybe we should look at why this confusion exists in the first place and whether we can address that instead. It's hard to suggest an approach here because I don't know exactly how this class is intended to be used, but an important question to that end would be "Why do we need a delegate and a callback at the same time"? Is there a design that gets rid of one? If not, can we shift things around to make it more obvious?

Similar to many of my comments before, making an architectural change here may not bring any meaningful impact for the particular PR at hand and I never try to force it. However, I do like bringing these up because I hope that it'll have an impact on how the developer will think this type of problem in the future and maybe come up with a solution that's clearer in future PRs. That's just my long way of saying, if you don't think it's worth discussing this, I am more than happy to drop it :)

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
oguzkocer: I think with this new implementation makes the following problem mentioned in the original comment more apparent:

> this little bit of code assumes knowledge of how something is implemented in another part of the library

Having said that, I think the only issue is the placement of the new method. So, how about we have a (private?) class with 2 methods for creating the `LogFile` and `storageURL`, is that maybe what you had in mind with `LogFileStore`?

```
private final class LogFileStore {
    func createLogFileFromURL(_ url: URL) -> LogFile {}
    func createStorageURLForLogFile(in: StorageDirectory, _ logFile: LogFile) -> String
}
```
_Wrote this on Github and although I can read Swift, that doesn't mean I can write it, so please imagine this more Swifty and with correct syntax 🙈_

It doesn't have to be a class, although I think there is some benefit to making it a class/protocol, the most important thing I was trying to convey here is to make these the bits of logic live together (possibly encapsulated) so we don't assume one in the other.

---

Now I am doubting if this discussion was really necessary 😬 @shiki Any thoughts on this discussion, am I maybe thinking way too hard about this?

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
oguzkocer: Perfect 🎉

Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
jkmassel: I like your point about

> Does the delegate require any information from the execution of the callback?

In this case, it does not – they're completely independent (but complimentary) tools.

In terms of 

> Why do we need a delegate and a callback at the same time?

The callback will be used to say "ok, uploading this file is finished, it's ok to remove this file from the upload queue", while the delegate allows the implementing application to respond to these fine-grained lifecycle events, if appropriate (consider a UI showing a list of files in the queue that need to be uploaded – it could be the delegate, and take appropriate action to update the UI based on what's happening).

Really appreciate your feedback :)

oguzkocer: In that case, I think this design is totally fine, thanks for explaining all that!

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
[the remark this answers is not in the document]
jkmassel: I went over this a bunch and tried with a protocol / implementation pair. While it works, it felt icky at this point since these classes are still `internal` and can't be overridden.

What I've done to simplify _this_ PR is to remove the `open` modifier on these classes. Because they can be imported into `XCTest` using `@testable`, which treats it as though it's `internal` to the test module,  it's effectively `internal open class`, which isn't useful to anyone.

Removing the `open` modifier has no effect on the tests, and if we need to make it `public` in a subsequent PR, I think it'll make sense to expose a `public` protocol with an `internal` default implementation at that point.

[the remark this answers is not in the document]
jkmassel: Oh – changed in a8fb2c03bf581e846288fa229bccfc26195265ed.

Automattic-Tracks-iOS/Services/EventLoggingNetworkService.swift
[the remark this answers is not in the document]
jkmassel: Good eyes – I adapted your suggestion for the comment in 1d9e9f886871f9b7b36fddf12eb2e551313d10d1.

jkmassel:
> It'd also be great if we can add comments to the public symbols like `EventLoggingDataSource` and `EventLoggingDelegate`. Comments like how they all connect together or how they should be used would be great. I suppose this could be in future iterations.

I agree – I'll do this as part of the next PR – the instructions on how to use them make more sense in the context of the wrapper class for this functionality, so I think it'd be hard to review here.

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
shiki: The way I understood this was that `EventLoggingUploadQueue` manages `LogFiles` that are inside a single directory. However, since a `LogFile` can be created by consumers, **there is no guarantee** that the `LogFile` passed to `remove()` is managed by `EventLoggingUploadQueue` itself. We currently handle that silently. 

I think you're right on the money here, @oguzkocer. I'd propose to even hide `Log.fileName` inside this `EventLoggingUploadQueue` as it is one of the queueing + storage implementation detail. 

But I find it challenging to give recommendations right now until I see how everything fits together — which I hope to see in upcoming PRs. Perhaps we could do an architectural review once all the classes have been put together?

Automattic-Tracks-iOS/Event Logging/EventLoggingDelegate.swift
shiki: 👍

Automattic-Tracks-iOS/Services/EventLoggingUploadQueue.swift
oguzkocer: > Perhaps we could do an architectural review once all the classes have been put together?

That sounds good to me 👍

Automattic-Tracks-iOSTests/Test Helpers/Mocks.swift
```diff
@@ -0,0 +1,107 @@
...
+@testable import AutomatticTracks
+
+typealias LogFileCallback = (LogFile) -> ()
+typealias ErrorWithLogFileCallback = (Error, LogFile) -> ()
+
+enum MockError: Error {
+    case generic
+}
+
+class MockEventLoggingDataSource: EventLoggingDataSource {
```
oguzkocer: I think the way this class is written is totally fine for tests, but here is how I would have written it myself: 872326118e4c1820192700f79a62d3f3de162bab.

The idea is to promote immutability and control how this class can be used. It has 0 value in this PR in practice, but I thought I'd take the time to mention this in a couple places in this PR for future reference, only if you find it useful of course.

Automattic-Tracks-iOSTests/Test Helpers/Mocks.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    func withEncryptionKeys() -> Self {
+        let keyPair = Sodium().box.keyPair()!
+        loggingEncryptionKey = Data(keyPair.publicKey).base64EncodedString()
+        return self
+    }
+}
+
+class MockEventLoggingDelegate: EventLoggingDelegate {
```
oguzkocer: Very similar to my last comment, here is how I'd have written this class: c3debf184004dddb662de750d4893d6c02f00c12

Automattic-Tracks-iOSTests/Tests/Event Logging/EventLoggingUploadManagerTests.swift
```diff
@@ -0,0 +1,76 @@
+import XCTest
+@testable import AutomatticTracks
+
+class EventLoggingUploadManagerTests: XCTestCase {
```
oguzkocer: In my previous PR review I tried to explain my thought process for property usage, but it was hard to do so without examples, so I took the time to create one for this PR: f7d87b2b11f2be9fe17344b9ef23f8f980f4ffae.

Similar to my previous PRs it doesn't have much value in tests, but I hope that it'll at least present a good example of how to work with immutability in tests.

There are a couple other files with very minor possible improvements similar to these, but I am leaving these because it's much of the same thing and mentioning these from here after won't have any example value either.

jkmassel:
@oguzkocer – I like the immutability suggestions for the tests – want to open another PR for those changes, and I can approve that one?

oguzkocer:
@jkmassel Looks like `develop` and that suggestion branch diverged quite a bit, so it creates a big merge conflict. Since the goal of that branch was mostly showcase a different way to go about things, and considering you're leaving for parental leave today :), I think it's fine to just let it die.

PR #136, opened 2020-06-17
Add encrypted logging upload authentication

jkmassel:
Adds support for the `Authorization` header with the data provided by the host app.

Reworks the test setup a bit in order to test the request creation in isolation.

### DIFF
--- a/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
@@ -4,11 +4,6 @@ import Sodium
 
 class EventLoggingUploadManager {
 
-    private enum Constants {
-        static let uuidHeaderKey = "log-uuid"
-        static let uploadHttpMethod = "POST"
-    }
-
     private let dataSource: EventLoggingDataSource
     private let delegate: EventLoggingDelegate
     private let networkService: EventLoggingNetworkService
@@ -34,17 +29,18 @@ class EventLoggingUploadManager {
             return
         }
 
-        guard let fileContents = fileManager.contents(atUrl: log.url) else {
+        guard fileManager.fileExistsAtURL(log.url) else {
             let error = EventLoggingFileUploadError.fileMissing
             delegate.uploadFailed(withError: error, forLog: log)
             callback(.failure(error))
             return
         }
 
-        var request = URLRequest(url: dataSource.logUploadURL)
-        request.addValue(log.uuid, forHTTPHeaderField: Constants.uuidHeaderKey)
-        request.httpMethod = Constants.uploadHttpMethod
-        request.httpBody = fileContents
+        let request = createRequest(
+            url: dataSource.logUploadURL,
+            uuid: log.uuid, authenticationToken:
+            dataSource.loggingAuthenticationToken
+        )
 
         delegate.didStartUploadingLog(log)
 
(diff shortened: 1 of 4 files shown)


Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -61,4 +57,13 @@ class EventLoggingUploadManager {
             }
         }
     }
+
+    func createRequest(url: URL, uuid: String, authenticationToken: String) -> URLRequest {
+        var request = URLRequest(url: url)
+        request.addValue(uuid, forHTTPHeaderField: "log-uuid")
```
jkmassel, line 63: By moving these values here, I don't anticipate we need them as constants anymore (if they were to change, the method wouldn't work anymore, and because the API is finalized, this should only change for a rev-bumped endpoint)

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -34,17 +29,18 @@ class EventLoggingUploadManager {
             return
         }
 
-        guard let fileContents = fileManager.contents(atUrl: log.url) else {
+        guard fileManager.fileExistsAtURL(log.url) else {
```
jkmassel, line 32: We don't need to read the file here – the API actually handles this, so it was a waste of resources – just validating that it exist allows the delegate callbacks to handle errors in the same way.

Automattic-Tracks-iOS/Event Logging/EventLoggingUploadManager.swift
```diff
@@ -34,17 +29,18 @@ class EventLoggingUploadManager {
...
             let error = EventLoggingFileUploadError.fileMissing
             delegate.uploadFailed(withError: error, forLog: log)
             callback(.failure(error))
             return
         }
 
-        var request = URLRequest(url: dataSource.logUploadURL)
```
jkmassel, line 44: Breaking this into a method allows it to be tested in isolation (which is why its no longer inlined)

Automattic-Tracks-iOSTests/Test Helpers/Mocks.swift
```diff
@@ -9,12 +9,13 @@ enum MockError: Error {
     case generic
 }
 
-struct MockEventLoggingDataSource: EventLoggingDataSource {
-    let loggingEncryptionKey: String
-    let previousSessionLogPath: URL?
-    var currentSessionLogPath: URL?
-    let logUploadURL: URL
-    let logUploadQueueStorageURL: URL
+class MockEventLoggingDataSource: EventLoggingDataSource {
```
jkmassel, line 12: This object was getting super heavy trying to do everything in the constructor. Using private setters effectively makes the object immutable, and the individual mutator methods are much more straightforward this way.

oguzkocer: I am not sure if I fully understand the issue here. The way we normally utilize this in Kotlin is to rely on the default values for the arguments and only pass what should be different for a test case. It looks like this would also work in Swift as such:
```
MockEventLoggingDataSource(logUploadUrl: url)
```

Am I missing something?

For the record, I am not against your current implementation, I just want to better understand the issue you're trying to solve.

jkmassel: Unfortunately, Swift doesn't support argument elision like this (and likely never will, though I can't find the post about that just now) – even though the arguments are named it's a more classical approach where they have to be in order. 

If I add a new method / property to the `EventLogginDataSource`, I have to put it in the mock constructor, which is quickly getting out of hand in tests (where I typically only want to customize one field at a time). That's why I figured I'd use this design - it keeps the `MockEventLoggingDataSource` code a lot shorter, as well as the tests (because they don't have to pass every argument if they don't need them)

PR #137, opened 2020-07-06
Simplify the Log File Interface

jkmassel:
This PR allows the hosting app to provide log files based on event severity and the timestamp of the event.

This should reduce the complexity of implementation in this library and in clients implementing it, while remaining flexible enough to handle the weirdness of different log file storage policies.

### DIFF
--- a/Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
@@ -38,10 +38,13 @@ extension EventLogging {
 
 extension Event {
     var errorType: EventLoggingErrorType {
-        if self.level == .fatal {
+        switch self.level {
+        case .fatal:
             return .fatal
+        case .error, .warning, .info, .debug:
+            return .debug
+        @unknown default:
+            return .debug
         }
-
-        return .debug
     }
 }
--- a/Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
+++ b/Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
@@ -1,6 +1,11 @@
 import Foundation
 import Sentry
 
+public enum EventLoggingErrorType {
+    case fatal
+    case debug
+}
+
 extension EventLogging {
     func attachLogToEventIfNeeded(event: Event) {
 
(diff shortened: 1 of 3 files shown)


Automattic-Tracks-iOSTests/Test Helpers/Mocks.swift
```diff
@@ -11,21 +11,24 @@ enum MockError: Error {
...
     init(
         encryptionKey: String = "foo",
         sessionLogPath: URL? = nil,
         logUploadUrl: URL = URL(string: "example.com")!,
         queueUrl: URL = FileManager.default.documentsDirectory.appendingPathComponent(UUID().uuidString)) {
-        loggingEncryptionKey = encryptionKey
-        previousSessionLogPath = sessionLogPath
-        logUploadURL = logUploadUrl
-        logUploadQueueStorageURL = queueUrl
+        self.loggingEncryptionKey = encryptionKey
```
jkmassel, line 24: This are all just changed to use `self` because I had to for sessionLogPath to disambiguate and I wanted it to be consistent.

Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
```diff
@@ -1,6 +1,11 @@
 import Foundation
 import Sentry
 
+public enum EventLoggingErrorType {
+    case fatal
+    case debug
```
oguzkocer, line 6: The name `Sentry` uses for this is `error`. Considering the implementation is domain agnostic, it's totally fine (maybe even better) to keep this naming scheme, but I wanted to mention this in case you had other thoughts.

Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
```diff
@@ -9,24 +14,34 @@ extension EventLogging {
...
+    }
+}
+
+extension Event {
+    var errorType: EventLoggingErrorType {
+        if self.level == .fatal {
+            return .fatal
+        }
+
+        return .debug
```
oguzkocer, line 45: I am not a fan of using `if-else` for enums because it both hides cases we currently handle and the cases that might be added after our initial implementation. The following might be more future-proof in that regard (although still not perfect), what do you think?

```suggestion
        switch self.level {
        case .fatal:
            return .fatal
        case .error, .warning, .info, .debug:
            return .debug
        @unknown default:
            return .debug
        }
```

Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
jkmassel: Which part are you referring to? The enum name, or one of the cases? :)

Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
jkmassel: I originally had this as a `switch` and actually changed it (for the same reason – I normally don't like `if` statements for enums). But in this case, I really only care about the `.fatal` cases being passed handled specially – anything else, regardless of what it is should be filed under `debug` (including if they were to, say, rename `.fatal` to `.crash` or add some new type here).

I don't mind the `switch`, though it's more verbose to do the same thing. There's nothing in our style guide to address this, but your suggestion is more formally correct, so let's do that 😄.

Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
oguzkocer: This was a comment on the `case debug` line. Github's UI in that regard is not super clear. I also got confused there for a second about which lines I added the comment for. The only difference between multiline comment and a single line comment seems to be the extra line: `Comment on lines 41 to 45`.

jkmassel: Ah yeah – I prefer `debug` in this case because it's not necessarily an error – we often send Sentry events for non-error states (like...this thing happened we didn't expect...)

Automattic-Tracks-iOS/Event Logging/EventLogging+Sentry.swift
oguzkocer: > But in this case, I really only care about the .fatal cases being passed handled specially

Although I knew that myself, I don't think that was 100% clear in the `if/else` version. A comment about the `if` line to indicate that probably would have been good enough, although it still wouldn't have been my personal preference.

PR #138, opened 2020-07-08
Fix encrypted logging race condition + add more logging

jkmassel:
This PR:
- Adds additional logging for issues that might affect encrypted log upload
- Adds logging infrastructure copied from other frameworks to support the logging efforts above
- Fixes a race condition in encrypted log upload


jkmassel:
>This mostly looks good to me, however some changes from #137 seems to have creeped into this PR. (some before we made changes to it) @jkmassel Could you let me know when you resolve the merge conflict for another look to make sure we don't merge stale changes?

I just rebased this PR on that one to make it impossible to merge stale changes – I branched before your last review, so this branch was missing your use of the `switch` statement. 

> Also, I did not review Automattic-Tracks-iOS/Extensions/CocoaLumberjack.swift which doesn't seem to be an implementation of this PR but rather a copy/move from another repo/place. Please let me know if I should review that as well.

This is perfect – I just copied it as-is. At some point it'd be great to fix this properly (using `cocoalumberjack/swift` directly), but I think that'll require a large amount of coordination between all of the dependencies to land the change in all of them at once.

PR #142, opened 2020-08-06
 Fix width and height tracking

aforcier:
Renames the height and width global params from `_ht` and `_wd` to `device_info_display_height` and `device_info_display_width` respectively.

I recently learned that the previous params were deprecated earlier this year, so they're currently not getting stored and historical data is also no longer available. The new ones are named consistently with Android and the general device info format.

Background: p4qSXL-3xm-p2

Corresponding Android PR: https://github.com/Automattic/Automattic-Tracks-Android/pull/48

I'm afraid I have absolutely no experience with updating pods so I humbly ask for help with getting this into the actual WPiOS app once merged. 🙏

### To test
1. Build WPiOS with this version of the Tracks library
2. Hook up a network listener like Charles, run WPAndroid, and wait for a tracks event to be broadcast (https://public-api.wordpress.com/rest/v1.1/tracks/record)
3. Examine the data sent, you should see a `commonProps` object which should include height/width information:
```js
"commonProps": {
  ...
  "device_info_display_height": "2160",
  "device_info_display_width": "1080",
  ...
```
9. Also confirm that the `_ht` and `_wd` props are no longer included

### DIFF
--- a/Automattic-Tracks-iOS/Services/TracksService.m
+++ b/Automattic-Tracks-iOS/Services/TracksService.m
@@ -40,8 +40,8 @@ NSString *const DeviceInfoOSVersionKey = @"device_info_os_version";
 NSString *const DeviceInfoBrandKey = @"device_info_brand";
 NSString *const DeviceInfoManufacturerKey = @"device_info_manufacturer";
 NSString *const DeviceInfoModelKey = @"device_info_model";
-NSString *const DeviceInfoHeightPixelsKey = @"device_info_display_height";
-NSString *const DeviceInfoWidthPixelsKey = @"device_info_display_width";
+NSString *const DeviceInfoHeightKey = @"device_info_display_height";
+NSString *const DeviceInfoWidthKey = @"device_info_display_width";
 NSString *const DeviceInfoNetworkOperatorKey = @"device_info_current_network_operator";
 NSString *const DeviceInfoRadioTypeKey = @"device_info_phone_radio_type";
 NSString *const DeviceInfoWiFiConnectedKey = @"device_info_wifi_connected";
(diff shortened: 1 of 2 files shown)


Automattic-Tracks-iOS/Services/TracksService.m
```diff
@@ -42,6 +40,8 @@ @interface TracksService ()
 NSString *const DeviceInfoBrandKey = @"device_info_brand";
 NSString *const DeviceInfoManufacturerKey = @"device_info_manufacturer";
 NSString *const DeviceInfoModelKey = @"device_info_model";
+NSString *const DeviceInfoHeightPixelsKey = @"device_info_display_height";
+NSString *const DeviceInfoWidthPixelsKey = @"device_info_display_width";
```
aerych, line 44: We could maybe remove "Pixels" from the consts since that had more to do with the (now obsolete) tracking pixel.

aerych:
> I'm afraid I have absolutely no experience with updating pods so I humbly ask for help with getting this into the actual WPiOS app once merged

No worries.  There's not much to it. 
First create a new [Release](https://github.com/Automattic/Automattic-Tracks-iOS/releases) for the beta version of the pod in this repo. Our CI tooling will take care of publishing a pod update. 
Next head over to the WPiOS and the WooCommerice-iOS repos.  Find the Podfile in each and update the version for the tracks pod. 
Normal PR review process for both apps. Ping me for reviews if you like. :)

Automattic-Tracks-iOS/Services/TracksService.m
aforcier: Oh interesting 🤔 I thought of it more as identifying the units (i.e. pixels and not density-independent pixels) than related to the tracking pixel. Regardless makes sense to simplify, I'll update.

aforcier: (Done in 054595b.)

PR #145, opened 2020-09-07
Remove nil coalescing for non-optional value in CrashLogging

mokagio:
The code generated this warning:

> Left side of nil coalescing operator '??' has non-optional type
> 'Bool', so the right side is never used

I noticed it while working on AutoProxxy.


loremattei (approved):
LGTM!

Issue #146, opened 2020-09-23
Where to store SQLite DB in non-sandboxed apps?


mokagio:
I'm working on adding support to an internal macOS app that is not sandboxed.

At launch I get prompted for access to the user's `~/Documents` folder.

I tracked it down to [`TracksContextManager` storing `.sqlite` DB at the URL for an `applicationDocumentsDirectory`](https://github.com/Automattic/Automattic-Tracks-iOS/blob/5f96537a62a8b7d3eb6635cb4dd8489dd39fc62a/Automattic-Tracks-iOS/Model/Core%20Data/TracksContextManager.m#L45-L57) computed as:

```objc
[[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory 
                                 inDomains:NSUserDomainMask] lastObject];
```

When running in a sandboxed app, like Simplenote for macOS, this path is 

```
~/Library/Containers/com.automattic.SimplenoteMac/Data/Documents
```

but for a non-sandboxed app, the path is `~/Documents`.

I think a good replacement would be the Application Support folder. Here's what the [macOS Library Directory Details](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html#//apple_ref/doc/uid/TP40010672-CH10-SW1) documentation says about it:

> Contains all app-specific data and support files. These are the files that your app creates and manages on behalf of the user and can include files that contain user data.

@jkmassel @jleandroperez what do you think?

mokagio:
I opened #147 to use Application Support instead of Documents and pointed the internal non-sandboxed app to it. It works fine now.

CI was failing since I integrated tracks due to a timeout because the app prompted for Documents access. The tests are now passing.

jleandroperez:
Yup @mokagio !! ApplicationSupport works super in these scenarios. Simperium uses that folder to keep track of "unsent changes".

[Ref. here!](https://github.com/Simperium/simperium-ios/blob/develop/Simperium/SPPersistentMutableDictionary.m#L399)

mokagio:
Excellent!

I'm not familiar with the way Tracks works. Do you know if there would be consequences in the Simplenote macOS behavior if we merged #147?

I imagine it may end up loosing all those events that haven't been log between the time the app restarted for the update and the new store being installed in Application Support.

I assume the impact would be negligible, but I'm not sure how to assess it.

[...]

PR #147, opened 2020-09-23
Add option to configure Tracks to store its DB in Application Support instead of Documents

mokagio:
See #146.

~~This is just a test to verify my internal non-sandboxed macOS app works fine when using Application Support.~~

Update: we decided to merge this change, but in a way that wouldn't change the behavior of Simplenote macOS.

To test these changes:

- I added unit tests 🙂, although CI is failing 🙃, but not because of my tests 🙂
- Checkout this PR in Simplenote macOS https://github.com/Automattic/simplenote-macos/pull/683

### DIFF
--- a/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
+++ b/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
@@ -3,8 +3,8 @@
 
 @interface TracksContextManager : NSObject
 
-@property (nonatomic, strong) NSManagedObjectContext *managedObjectContext;
-@property (nonatomic, strong) NSManagedObjectModel *managedObjectModel;
-@property (nonatomic, strong) NSPersistentStoreCoordinator *persistentStoreCoordinator;
+@property (nonnull, nonatomic, strong) NSManagedObjectContext *managedObjectContext;
+@property (nonnull, nonatomic, strong) NSManagedObjectModel *managedObjectModel;
+@property (nonnull, nonatomic, strong) NSPersistentStoreCoordinator *persistentStoreCoordinator;
 
 @end
--- a/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
+++ b/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
@@ -7,4 +7,18 @@
 @property (nonnull, nonatomic, strong) NSManagedObjectModel *managedObjectModel;
 @property (nonnull, nonatomic, strong) NSPersistentStoreCoordinator *persistentStoreCoordinator;
 
+/// In version 0.5.1 and earlier, the CoreData store was always saved in the Documents folder.
+///
+/// This approach is fine for sandboxed applications distributed via the Mac App Store, but would
+/// result in the user being asked for access to the `~/Documents` folder in non-sandboxed apps.
+///
+/// To avoid this bad UX – which also blocks tests in CI with the same alert – when the app is
+/// sandboxed, we can save the store in `~/Application Support`.
+///
+/// This designated initializer allows consumers to tell `TracksContextManager` the mode in which
+/// the app is running and where to save the store.
+///
+/// The `super` `init` method defaults to `false` for backward compatibility.
+- (nonnull instancetype)initWithSandboxedMode:(BOOL)sandboxed NS_DESIGNATED_INITIALIZER;
+
 @end
--- a/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
+++ b/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
@@ -42,7 +42,7 @@
     // Create the coordinator and store
     
     _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:[self managedObjectModel]];
-    NSURL *storeURL = [[self applicationSupportURLForContainerApp] URLByAppendingPathComponent:@"Tracks.sqlite"];
+    NSURL *storeURL = [self storeURL];
     NSError *error = nil;
     if (![_persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType configuration:nil URL:storeURL options:nil error:&error]) {
         
@@ -57,6 +57,11 @@
     }
     
     return _persistentStoreCoordinator;
+}
+
+- (NSURL *)storeURL {
+    NSURL *containerDirectoryURL = [self applicationSupportURLForContainerApp];
+    return [containerDirectoryURL URLByAppendingPathComponent:@"Tracks.sqlite"];
 }
 
 - (NSURL *)applicationSupportURLForContainerApp {
--- a/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
+++ b/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
@@ -10,7 +10,7 @@
 @implementation TracksContextManager
 
 - (instancetype)init {
-    return [[TracksContextManager alloc] initWithSandboxedMode:true];
+    return [self initWithSandboxedMode:true];
 }
 
 - (instancetype)initWithSandboxedMode:(BOOL)sandboxed {
--- a/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
+++ b/Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
@@ -42,7 +42,7 @@
     // Create the coordinator and store
     
     _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:[self managedObjectModel]];
-    NSURL *storeURL = [[self applicationDocumentsDirectory] URLByAppendingPathComponent:@"Tracks.sqlite"];
+    NSURL *storeURL = [[self applicationSupportURLForContainerApp] URLByAppendingPathComponent:@"Tracks.sqlite"];
     NSError *error = nil;
     if (![_persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType configuration:nil URL:storeURL options:nil error:&error]) {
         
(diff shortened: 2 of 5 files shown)


Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
```diff
@@ -3,8 +3,22 @@
 
 @interface TracksContextManager : NSObject
 
-@property (nonatomic, strong) NSManagedObjectContext *managedObjectContext;
-@property (nonatomic, strong) NSManagedObjectModel *managedObjectModel;
-@property (nonatomic, strong) NSPersistentStoreCoordinator *persistentStoreCoordinator;
+@property (nonnull, nonatomic, strong) NSManagedObjectContext *managedObjectContext;
+@property (nonnull, nonatomic, strong) NSManagedObjectModel *managedObjectModel;
+@property (nonnull, nonatomic, strong) NSPersistentStoreCoordinator *persistentStoreCoordinator;
```
mokagio, line 8: I had to add the annotation here too to avoid warnings. It makes the diff a bit noisier but it's good to have it anyways.

The implementation of all those getters makes sure they are not `nil`, or rather, fails if they are, so `nonnullable` seemed appropriate.

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
```diff
@@ -3,8 +3,22 @@
...
+/// result in the user being asked for access to the `~/Documents` folder in non-sandboxed apps.
+///
+/// To avoid this bad UX – which also blocks tests in CI with the same alert – when the app is
+/// sandboxed, we can save the store in `~/Application Support`.
+///
+/// This designated initializer allows consumers to tell `TracksContextManager` the mode in which
+/// the app is running and where to save the store.
+///
+/// The `super` `init` method defaults to `false` for backward compatibility.
+- (nonnull instancetype)initWithSandboxedMode:(BOOL)sandboxed NS_DESIGNATED_INITIALIZER;
```
mokagio, line 22: I wonder whether the parameter should be an enum instead:

```swift
_ = TracksContextManager(storeLocation: .applicationSupport)
// or
_ = TracksContextManager(storeLocation: .documents)
```

I'm on the fence and might sit overnight. Regardless, because the changes doesn't affect existent consumers thanks to the `init` override below, I think it's okay to merge this as it is and then I can follow up with a refactor if we think it's worth it.

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
```diff
@@ -3,10 +3,25 @@
...
 
 @implementation TracksContextManager
 
+- (instancetype)init {
+    return [[TracksContextManager alloc] initWithSandboxedMode:true];
+}
+
+- (instancetype)initWithSandboxedMode:(BOOL)sandboxed {
+    self = [super init];
+    if (!self) { return nil; }
```
mokagio, line 18: This should never happen so I defined the initializer as `nonnullable`. In Swift land, that would be the case.

mokagio (commented):
I refined the implementation to avoid changing the way existing consumers of the library store the DB.

@jleandroperez @jkmassel can I ask you for a new round of review?

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
jleandroperez: Yup that's the perfect approach Gio!

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
jleandroperez: To be **entirely** honest, I would keep the initializer simple, and just switch to the Application Support folder.

Something in the lines of:

- If a `Documents` folder already exists, use it
- Otherwise, check if a folder in `Applications Support` exists, and use it
- If none of the above, create a new folder in `Applications Support`

WYDT?

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
jleandroperez: Standard ObjC, exactly!

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
```diff
@@ -3,10 +3,25 @@
...
 @interface TracksContextManager ()
 
+@property (nonatomic, assign) BOOL sandboxed;
+
 @end
 
 @implementation TracksContextManager
 
+- (instancetype)init {
+    return [[TracksContextManager alloc] initWithSandboxedMode:true];
```
jleandroperez: This initializer returning a new instance, rather than the one being initialized.

You would typically instantiate a Context Manager like this:

```
TracksContextManager *manager = [[TracksContextManager alloc] init];
```

1. `alloc` reserves memory for the object
2. `init` is expected to initialize the instance

In this case, `init` is alloc'ating more memory, and returning yet another instance.

jleandroperez:
```suggestion
    return [self initWithSandboxedMode:true];
```

jleandroperez: (Sending you a change suggestion to address that!)

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
```diff
@@ -59,6 +74,64 @@ - (NSPersistentStoreCoordinator *)persistentStoreCoordinator {
...
+                                                   error:&error];
+
+    // It seems safe not to handle this error because Application Support should always be
+    // available.
+    if (error != nil) {
+        DDLogError(@"Failed to get path to Application Support folder: %@, %@", error, [error userInfo]);
+        abort();
+    }
+
+    return applicationSupportURL;
```
jleandroperez: You can proooobably simplify this API as follows:

```
NSURL *appSupportURL = [[[NSFileManager defaultManager] URLsForDirectory:NSApplicationSupportDirectory inDomains:NSUserDomainMask] lastObject];
```

[Ref. here](https://github.com/Simperium/simperium-ios/blob/develop/Simperium/SPPersistentMutableDictionary.m#L400)

(The Application Support path is expected to always be there!)

jleandroperez (changes requested):
@mokagio sir!! Few changes suggested:

- Initializer fix (would prevent leaking memory)
- ApplicationSupport API simplification (not required, not a blocker either!!)

As per the initializer detail, 10000% up to you!

- Taking that parameter isn't a problem at all
- It can be solved without specifying if we're sandboxed or not, as well. Yet another alternative, is to just always use the Application Support path, and [migrate legacy databases, if needed](https://github.com/Simperium/simperium-ios/blob/develop/Simperium/SPPersistentMutableDictionary.m#L374)

I would take the easiest path. As usual 😄

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.h
jkmassel: I'm ok with either approach. I like how pretty the enum is, but I like how much work @jleandroperez's isn't 😄

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
mokagio: Thank you for this! Gosh 😳 I'm really rusty at writing Objective-C 🤦 

I felt like there was something wrong with that, but it compiled and the tests passed so I shrugged it off.

Interestingly, Xcode also gives me a warning about it

image

It either came after I wrote the code and moved on or after I added the `NS_DESIGNATED_INITIALIZER` annotation. Regardless, I shouldn't have missed it... 😳

mokagio:
Thanks @jleandroperez for the great review!

Regarding the init, I like the idea of keeping it simple and eventually migrate everything to Application Support.

I haven't tried, but I think that even to just check whether a file exists in the Documents folder from a non-sandboxed app, macOS would request the user for permission. If that's the case, then we'd be back to square one, because the real aim of this change is avoid that behavior in non-sandboxed apps.

I really like the migration idea, but I think it needs more experimenting and might be best for a dedicated PR.

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
```diff
@@ -59,6 +74,51 @@ - (NSPersistentStoreCoordinator *)persistentStoreCoordinator {
...
+- (NSURL *)storeURL {
+    return [[self storeContainerDirectoryURL] URLByAppendingPathComponent:@"Tracks.sqlite"];
+}
+
+- (NSURL *)storeContainerDirectoryURL {
+    if (self.sandboxed == YES) {
+        return [self applicationSupportURLForContainerApp];
+    } else {
+        return [self applicationDocumentsDirectory];
+    }
```
jleandroperez, line 86: @mokagio Is it possible this logic is inverted? (sandboxed > Documents Directory?).

Otherwise non sandboxed apps would still trigger the Documents Folder alert perhaps?

jleandroperez (approved):
@mokagio One tiny comment sent away (apologies I missed that!!).

Other than that, please :shipit: when ready!!

Automattic-Tracks-iOS/Model/Core Data/TracksContextManager.m
mokagio: You are totally right!

I didn't notice because the tests were inverted, too 🤦‍♂️

mokagio:
😫 I feel like I'm in a deadlock... There's some inconsistent issue with the tests, which [doesn't happen with the Xcode 12](https://github.com/Automattic/Automattic-Tracks-iOS/pull/150), but I can't upgrade to Xcode 12 because [the pod validation fails](https://github.com/Automattic/Automattic-Tracks-iOS/issues/149).

Inconsistent as in [this](https://app.circleci.com/pipelines/github/Automattic/Automattic-Tracks-iOS?branch=store-data-in-application-support):

image

jleandroperez (approved):
:shipit: sir!!

jleandroperez:
For the record, Cocoapods had issues finding the Sodium-Fork framework. Workaround involved running `bundle exec pod repo update`

Thank you Gio!!

PR #151, opened 2020-10-12
Bump version to 0.6.0

mokagio:
If I understand the workflow correctly, once this is merged I should tag the merge commit and CI will do the rest, i.e: publish to CoacoPods trunk.


jkmassel (approved):
> If I understand the workflow correctly, once this is merged I should tag the merge commit and CI will do the rest, i.e: publish to CoacoPods trunk.

Mostly – you'll want to land this in `trunk` then tag there, but the rest is accurate :)

PR #152, opened 2020-10-13
Add timestamps to new events

jkmassel:
This fixes a potential crash if these events are later used with Encrypted Logging


shiki (approved):
:shipit: Thanks!

PR #155, opened 2020-11-06
Adds AB Testing support (ExPlat)

leandroalonso:
This PR adds AB Testing support.

### Configuration

(WPiOS PR here: https://github.com/wordpress-mobile/WordPress-iOS/pull/15236)

Before having fun with AB testing, the app needs to configure the ExPlat instance. To do so, the `switchToAuthenticatedUserWithUsername` method was changed and now it requires a `token` parameter.

An example from WPiOS:

```swift
[self.tracksService switchToAuthenticatedUserWithUsername:username userID:@"" token:self.token skipAliasEventCreation:NO];
```

Then, you need to call the method `refreshIfNeeded()` in `didFinishLaunchingWithOptions`.

```swift
ExPlat.shared.refreshIfNeeded()
```

And that's all. :)

### Usage

Once it's configured you can use like this:

```swift
switch ExPlat.shared.experiment("experiment-name") {
    case .treatment:
        // Show treatment option
    case .control:
        // Show control option
    case .unknown:
        // Experiment doesn't exist or wasn't retrieved
    case .other(let value) where value == "my-value":
        // any other value 
}
```

Or, more simpler:

```swift
switch ExPlat.shared.experiment("experiment-name") {
    case .treatment:
        // Show treatment option
    default:
        // Show control option
}
```

### DIFF
--- a/Automattic-Tracks-iOS/ABTesting/ExPlat.swift
+++ b/Automattic-Tracks-iOS/ABTesting/ExPlat.swift
@@ -1,7 +1,7 @@
 import Foundation
 import UIKit
 
-public class ExPlat: ABTesting {
+@objc public class ExPlat: NSObject, ABTesting {
     public static var shared: ExPlat!
 
     private let service: ExPlatService
@@ -22,8 +22,19 @@ public class ExPlat: ABTesting {
     public init(configuration: ExPlatConfiguration,
          service: ExPlatService? = nil) {
         self.service = service ?? ExPlatService(configuration: configuration)
+        super.init()
         subscribeToNotifications()
         ExPlat.shared = self
+    }
+
+    @objc public static func configure(platform: String,
+                      oAuthToken: String?,
+                      userAgent: String?,
+                      anonId: String?) {
+        _ = ExPlat(configuration: ExPlatConfiguration(platform: platform,
+                                                       oAuthToken: oAuthToken,
+                                                       userAgent: userAgent,
+                                                       anonId: anonId))
     }
 
     deinit {
--- a/Automattic-Tracks-iOS/ABTesting/ExPlat.swift
+++ b/Automattic-Tracks-iOS/ABTesting/ExPlat.swift
@@ -1,0 +1,129 @@
+import Foundation
+import UIKit
+
+public class ExPlat: ABTesting {
+    public static var shared: ExPlat!
+
+    private let service: ExPlatService
+
+    private let assignmentsKey = "ab-testing-assignments"
+    private let ttlDateKey = "ab-testing-ttl-date"
+
+    private var ttl: TimeInterval {
+        guard let ttlDate = UserDefaults.standard.object(forKey: ttlDateKey) as? Date else {
+            return 0
+        }
+
+        return ttlDate.timeIntervalSinceReferenceDate - Date().timeIntervalSinceReferenceDate
+    }
+
+    private(set) var scheduledTimer: Timer?
+
+    public init(configuration: ExPlatConfiguration,
+         service: ExPlatService? = nil) {
+        self.service = service ?? ExPlatService(configuration: configuration)
+        subscribeToNotifications()
+        ExPlat.shared = self
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
+        NotificationCenter.default.removeObserver(self, name: UIApplication.willEnterForegroundNotification, object: nil)
+    }
+
+    /// Only refresh if the TTL has expired
+    ///
+    public func refreshIfNeeded(completion: (() -> Void)? = nil) {
+        guard ttl > 0 else {
+            completion?()
+            scheduleRefresh()
+            return
+        }
+
+        refresh(completion: completion)
+    }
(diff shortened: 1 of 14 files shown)


leandroalonso:
@mokagio I removed your private specs because the build was failing on the CI machine. Just want to double-check if this will not cause any unintended issue.

astralbodies (commented):
It probably makes sense to tie the ExPlat service into the Tracks service somehow so it can respond to auth/de-auth scenarios already set up in the host app.

Methods used by the host app to track auth: https://github.com/Automattic/Automattic-Tracks-iOS/blob/trunk/Automattic-Tracks-iOS/Services/TracksService.h#L29

Those two methods are already hooked in somewhere in the host app based on the auth lifecycle. The only thing this doesn't give you is the OAuth bearer token, of course. Just figured since we already deal with going from anon -> logged in -> anon with Tracks.

astralbodies:
Also, consider adding the OAuth Bearer token to the methods tracking authentication state. Since that token changes at the same time (usually), the TracksService could be the owner of that data, and then later inject that into the ExPlat service for use.

Does ExPlat work without a Bearer token? Wonder how that affects anonymous users / users not yet logged in.

mokagio:
Hey @leandroalonso, sorry for the long time it took me to get to this 😞 🙏 

You did well. I'm pretty sure I added them in there only because the CDN didn't seem to find the `.podspec` for my `Sodium-Fork` for some reason (update propagation maybe?). The error we got in CI clearly shows that the spec is now there, so good to clean it 🧹

leandroalonso:
@astralbodies ready for another review!

ExPlat is now coupled with `Tracks` so it doesn't need to be configured, but it needs to be started in `didFinishLaunchingWithOptions`.

I have one question: I kept both the old method and the new one with a `token` param. My intention is to not break the apps that use this library. So, if the app wants to use the AB testing feature it needs to switch to the `switchToAuthenticatedUserWithUsername` with the `token` param. Thoughts about that?

astralbodies:
@leandroalonso I'm terribly sorry for missing this code review. I will be reviewing it today!

Automattic-Tracks-iOS/Services/TracksService.h
```diff
@@ -27,6 +27,7 @@ extern NSString *const TrackServiceDidSendQueuedEventsNotification;
 - (NSDictionary *)dictionaryForTracksEvent:(TracksEvent *)tracksEvent withParentCommonProperties:(NSDictionary *)parentCommonProperties;
 
 - (void)switchToAuthenticatedUserWithUsername:(NSString *)username userID:(NSString *)userID skipAliasEventCreation:(BOOL)skipEvent;
```
astralbodies, line 29: I was thinking about maybe having you add a deprecation warning here... But then I think Simplenote uses this client as well, and they don't use WordPress.com logins. Now that has me wondering if we should name the `token` in the new method to be a wpcom bearer token for clarity.

leandroalonso: Do you mean something like that?

```swift
- (void)switchToAuthenticatedUserWithUsername:(NSString *)username userID:(NSString *)userID wpComToken:(NSString *)token skipAliasEventCreation:(BOOL)skipEvent;
```

astralbodies (approved):
This looks good! I didn't test out the functionality with the backend system. I did suggest maybe changing the parameter name a bit for clarity's sake.

Automattic-Tracks-iOS/Services/TracksService.h
astralbodies: Yup, like that!

leandroalonso: Done in 7f8c48411016516e9fe6f7862e693e15199b41dc

astralbodies:
@leandroalonso Ah crap, I didn't notice this PR was targeting `trunk`.

@jkmassel can we fix that?

leandroalonso:
@astralbodies I already fixed. 😅

Automattic-Tracks-iOS/ABTesting/ExPlat.swift
```diff
@@ -0,0 +1,140 @@
...
+            self.scheduleRefresh()
+
+            completion?()
+        }
+    }
+
+    public func experiment(_ name: String) -> Variation {
+        guard let assignments = UserDefaults.standard.object(forKey: assignmentsKey) as? [String: String?],
+              case let variation?? = assignments[name] else {
+            return .unknown
```
withinboredom, line 82: It may be better to consider returning `.control` instead of `.unknown` so that an old client can properly fall back to the control experience if the experiment is no longer available.

Automattic-Tracks-iOSTests/Mock Data/explat-assignments.json
```diff
@@ -0,0 +1,6 @@
+{
+    "variations": {
+        "experiment": "control"
```
withinboredom, line 3: Explat will never return the name of the control variation, For control, it always returns `null` and/or it won't exist in the response. It only returns the name of treatment variations. Right now, we only support a singular treatment, but multi-variate experiments are in the works.

PR #131, opened 2020-05-20
Update the Sentry SDK to 5.x

jkmassel:
- Updates the Sentry SDK without causing any breaking changes
- Fixes up tests to be compatible with the new SDK (some weren’t particularly useful)
- Adds a switch to force-enable Sentry events in debug builds (`-force-crash-logging=1`)

PR #157, opened 2020-12-09
Switch back to Sodium's original library & update to 0.9.0

mokagio:
This is a demo PR with a version of Tracks that builds for macOS targets on Xcode 12.2.

I don't consider it ready for merge because of the huge dependency of updating Tracks to version 5.0. Something on which @jkmassel also worked on (#131) but that I didn't realize when I worked on this.

It's handy to have the code online so that we can point to this branch in our macOS targets and use this PR for Sodium related discussions.


mokagio:
The validation fails **with warnings only**.

mokagio:
Closing in favor of #166 and #163.

PR #159, opened 2020-12-17
Update Sentry (adding in manual event sending)

jkmassel:

PR #160, opened 2021-01-11
Updates Sodium to 0.9.1 and removes workaround for M1 builds

astralbodies:
This PR is an attempt to get the Tracks library to build properly on M1 devices. The following updates were made:

* Cocoapods 1.10.0 - Since this is only a local change to the project, I figured why not. Doesn't affect the podspec.
* Sodium - bumped to 0.9.1 which includes fixes for building on Xcode 12 and on M1 architecture machines.
* Deployment version - Made 12.0 the minimum to support the changes inside of Sodium.
* TracksDemo - refreshed the dependencies and got it to run/compile.

I did not update the podspec version as it's been forever since I've done a release for a pod.


Automattic-Tracks-iOS.podspec
```diff
@@ -32,24 +32,5 @@ Pod::Spec.new do |spec|
...
-
-  # Notice that a CocoaPods contributor discourages the use of
-  # `user_target_xcconfig`, but in this case also agrees that there might not be
-  # a better approach.
-  # See
-  # https://github.com/CocoaPods/CocoaPods/issues/10065#issuecomment-701055569
-  spec.user_target_xcconfig = {
-    'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64',
-  }
+  spec.dependency 'Sodium', '0.9.1'
```
mokagio: Excellent! I missed the 0.9.1 update with the arm64 update 🎉

Gemfile
```diff
@@ -1,4 +1,6 @@
-source "https://rubygems.org" do
-  gem "cocoapods", '~> 1.8.0'
-  gem "xcpretty"
-end
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+gem 'cocoapods', '~> 1.10.0'
```
mokagio: Should relax this a bit and go for `~> 1.10`? Not a requirement for this PR, though, just a question.

TracksDemo/TracksDemo.xcodeproj/project.pbxproj
```diff
@@ -385,21 +389,21 @@
...
-			inputFileListPaths = (
-			);
 			inputPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-TracksDemo Mac/Pods-TracksDemo Mac-frameworks.sh",
 				"${BUILT_PRODUCTS_DIR}/Automattic-Tracks-iOS-macOS/AutomatticTracks.framework",
 				"${BUILT_PRODUCTS_DIR}/CocoaLumberjack-macOS/CocoaLumberjack.framework",
 				"${BUILT_PRODUCTS_DIR}/Reachability-macOS/Reachability.framework",
+				"${BUILT_PRODUCTS_DIR}/Sentry-macOS/Sentry.framework",
+				"${BUILT_PRODUCTS_DIR}/Sodium-macOS/Sodium.framework",
 			);
```
mokagio, line 399: Side note and future improvement reminder. Not required for this PR.

Starting 1.10.0, CocoaPods uses the Xcode project `compatibilityVersion` value to decide whether to use a file with the list of frameworks or enumerate them in the project. See [this PR](https://github.com/CocoaPods/CocoaPods/commit/d4e8141261e40e3e2d9a201d627457a2c11eb545) and [this SO question](https://stackoverflow.com/questions/64254389/cocoapods-1-10-beta-2-unable-to-load-contents-of-file-list-xcfilelist).

Using the file list is better because it replaces this kind of noisy diff with a change in a dedicated file. To support it, we shall upgrade the project file to the Xcode 11.0 compatible format.

image

mokagio (commented):
Thank you so much @astralbodies and @jkmassel for this 😍

The code looks good. I'll fetch it on Simplenote macOS and our internal macOS app later to test it properly. Just wanted to leave this thank you note in the meantime.

mokagio:
🤔 I tried this on Simplenote macOS and our internal macOS app. On Simplenote, it archives fine, but the resulting archive is only for Intel

image

On our internal app, it builds fine but fails to archive. I believe the issue has to do with arm64 for two reasons:

One: if I tell Xcode to "Build Active Architecture Only" in the Release configuration (the one used by the Archive action), archiving succeeds

image

Two: The build error I get happens in what looks like the compilation for arm64

image

I pushed the branches I used to test this, in both repos they're called [`test-tracks-m1-pr`](https://github.com/Automattic/Simplenote-macos/tree/test-tracks-m1-pr).

Automattic-Tracks-iOS.podspec
jkmassel: This should be `~> 0.9` – the `.lock` file will ensure we never downgrade (unless explicitly asked to), and this way other dependencies that may use sodium won't break.

Gemfile
jkmassel: Yep, though I disagree – it should be a requirement for this PR please!!! 😄

Podfile
```diff
@@ -8,11 +8,11 @@ def shared
   pod 'CocoaLumberjack', '~> 3'
   pod 'Reachability', '~> 3'
   pod 'Sentry', '~> 4'
-  pod 'Sodium-Fork', '~> 0.8.2'
+  pod 'Sodium', '~> 0.9.1'
```
jkmassel: This could be `~> 0.9` as well – CocoaPods won't downgrade the version.

TracksDemo/TracksDemo.xcodeproj/project.pbxproj
jkmassel: I agree that the file list would be super nice!!! This one I think is a great fit for a future PR.

jkmassel (changes requested):

Automattic-Tracks-iOS.podspec
astralbodies: Since `0.9.0` won't work for us we should in theory specify `0.9.1` because a podspec can't enforce a hash like a lock file can. If someone doesn't update the Cocoapods repo and 0.9.0 is the latest version, then that's the one that'll get installed. If they follow semantic versioning, it's effectively doing the same thing as `~> 0.9` because there is no version `1.10.x`.

astralbodies: (`0.9.1` has the arm64 fix)

Gemfile
astralbodies: I'll change it!

Automattic-Tracks-iOS.podspec
jkmassel: Ah – you are correct – the `.lock` file only applies to a `podfile` 🤦 

In this case it would be preferable to use `>= 0.9.1` IMHO – no lower, but allow higher versions if they're used in other dependencies to avoid conflicts.

astralbodies: Ooh, yes, that's a better one!

mokagio: > In this case it would be preferable to use >= 0.9.1 IMHO – no lower, but allow higher versions if they're used in other dependencies to avoid conflicts.

+1 👍

Podfile
astralbodies: Agreed to `>= 0.9.1` in Slack.

jkmassel (approved):
LGTM! Merging to start rebasing https://github.com/Automattic/Automattic-Tracks-iOS/pull/159

PR #162, opened 2021-01-12
Merge AB Testing support into develop (ExPlat)

leandroalonso:
Merge https://github.com/Automattic/Automattic-Tracks-iOS/pull/155 to the correct branch (not `trunk`)


jkmassel (approved):
The commit hashes for the changes are identical, meaning the code is the same. The only difference is the merges, which are clean and just fast-forward the history, so no concerns about merging this.

:shipit:

PR #164, opened 2021-01-13
Add macOS support for ExPlat

jkmassel:
While working on #159 I noticed that the `ABTesting` code wasn't targeting macOS. Because we want to keep this library in sync across both platforms, this PR just runs around and adds everything to the macOS target as well – it was pretty much all already compatible! 😅 

**To Test:**
- Note that we're now running the ExPlat tests with our macOS tests
- Note that the tests pass!

### DIFF
--- a/Automattic-Tracks-iOS.xcodeproj/project.pbxproj
+++ b/Automattic-Tracks-iOS.xcodeproj/project.pbxproj
@@ -7,6 +7,14 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


leandroalonso (approved):
@jkmassel I guess we can remove the last exclude path from this line: https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Automattic-Tracks-iOS.podspec#L15

Then we're good to go! Thanks for this improvement. :)

jkmassel:
>I guess we can remove the last exclude path from this line

Yeah, good call thanks!!

jkmassel:
So after trying that, there's a problem with updating to macOS 10.12 (because it starts building for arm64 at that point, which isn't supported by Sentry).

We'll need to land #159 first, then remove that chunk of the `podspec`.

To avoid conflicts, I'll merge this for now and fix up the `podspec` once that work is done if that's ok?

PR #166, opened 2021-01-16
Update Sentry to v6

jkmassel:
Updates the Sentry library to v6, which enables:
- Session tracking correlated with crashes to give us better crash rate data
- Support for building the project on arm64 macs

Unfortunately, the v6 library takes away our ability to manually send events to Sentry, so this PR adds that ability back by using the API directly.

Previously to this PR, testing Sentry integration was always a bit tricky because we'd need to integrate into a host app and make weird changes to it in order to test the integration.

This PR adds the ability to test Sentry functionality from the example project.

**To Test:**
- `cd` into the `TracksDemo` directory and run `bundle exec pod install`
- Create the `Shared/Secrets.swift` file:
   - For automatticians: copy the file from our mobile secrets repo (under `iOS/tracks-demo`). You might need to update (`git pull`) the mobile secrets repo first.
   - For non-automatticians: copy the file `Shared/Secrets.example.swift` and fill in the missing details
   - Feel free to customize your email address :)
   - _Note: In a future PR, we'll integrate the `configure` functionality to automate this, but it's not part of this PR in order to avoid making it any larger._
- Run the sample project on iOS. Go to the "Crashes" tab and try "Send Test Event" and "Send Error and Wait". Ensure that both events show up in the Sentry `tracks-demo` project with an associated stack trace.
- Run the same test on the macOS project. 

**Additional M1-specific tests:**
- Try building for `Any iOS Device`
- Try building for `My Mac`, and `Any Mac (Apple Silicon, Intel)` [ note: you may need to clean your build directory first – this tends to fail otherwise – I assume this is an Xcode bug ]

_I'd love one review on this from a mobile infrastructure engineer as well as a review from someone who has M1 hardware to test with._


Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
 import Sentry
 import CocoaLumberjack
 
+struct CrashLoggingInternals {
```
jkmassel: This class isn't fully untangled from all the `static` business, so we keep an internal-only static reference around until we can get rid of it.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+    static var crashLogging: CrashLogging?
+}
+
 /// A class that provides support for logging crashes. Not compatible with Objective-C.
 public class CrashLogging {
 
-    /// A singleton is maintained, but the host application needn't be aware of its existence.
```
jkmassel, line 8: No more static instance internally – makes things safer, easier to test, and lets us remove a bunch of code.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
-    /**
-     Initializes the crash logging system.
-
-     - Parameters:
-     - dataProvider: An object that will provide any required data to the crash logging system.
-
-     - SeeAlso: CrashLoggingDataProvider
-     */
-    public static func start(withDataProvider dataProvider: CrashLoggingDataProvider, eventLogging: EventLogging? = nil) {
+    public func start() throws -> CrashLogging {
```
jkmassel, line 29: `throws` because the DSN may be invalid, and we don't want to hide that error – otherwise we might ship an app to production with no ability to track crashes

Automattic-Tracks-iOS.podspec
```diff
@@ -23,14 +23,14 @@ Pod::Spec.new do |spec|
...
+  spec.osx.deployment_target  = '10.12'
 
   spec.header_dir = 'AutomatticTracks'
   spec.module_name = 'AutomatticTracks'
 
   spec.ios.dependency 'UIDeviceIdentifier', '~> 1'
   spec.dependency 'CocoaLumberjack', '~> 3'
   spec.dependency 'Reachability', '~> 3'
-  spec.dependency 'Sentry', '~>4'
+  spec.dependency 'Sentry', '~> 6'
```
mokagio, line 34: Thank you for adding _the space_, too 😄

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
mokagio: I like the idea of shoving all of the stuff that we cannot easily migrate into a dedicated place to make it easier to find and change or remove later 👍 

I'm addicted to nested types recently. I can see this fit nicely as one

```swift
public class CrashLogging {

  /// Comment explaining why this is here...
  class Internals {
     // ...
  }

  // ...
  Internals.crashLogging = self
  // ...
}
```

No need to implement it if you don't see the benefit. Just playing code golf.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
 
-        // Only allow initializing this system once
-        guard !isStarted else { return }
-        isStarted = true
+        /// Validate the DSN ourselves before initializing, because the SentrySDK silently prints the error to the log instead of telling us if the DSN is valid
```
mokagio, line 31:
```suggestion
        // Validate the DSN ourselves before initializing, because the SentrySDK silently prints the error to the log instead of telling us if the DSN is valid
```

I'm suggesting to go from `///` to `//` because the former is to be used for documentation comments, but this one is not an header doc of a type or method.

mokagio: > the Sentry SDK silently prints the error to the log instead of telling us if the DSN is valid

Have you considered opening an issue with a feature suggestion for this.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
-            if dataProvider.shouldEnableAutomaticBreadcrumbTracking {
-                Client.shared?.breadcrumbs.maxBreadcrumbs = 500 // Store lots of breadcrumbs to trace errors
-                Client.shared?.enableAutomaticBreadcrumbTracking()
-            }
+            options.beforeSend = self.beforeSend
 
-            // Override event serialization to append the logs, if needed
-            Client.shared?.beforeSerializeEvent = sharedInstance.beforeSerializeEvent
-            Client.shared?.shouldSendEvent = sharedInstance.shouldSendEvent
+            /// Attach stack traces to non-fatal errors
```
mokagio, line 44:
```suggestion
            // Attach stack traces to non-fatal errors
```

Same rationale as above.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+            event.tags = [String: String]()
         }
-    }
 
-    /// A Sentry hook that controls whether or not the event should be sent.
-    private func shouldSendEvent(_ event: Event?) -> Bool {
+        event.tags?["locale"] = NSLocale.current.languageCode
 
-        DDLogDebug("📜 Firing `shouldSendEvent`")
+        /// Always provide a value in order to determine how often we're unable to retrieve the value
```
mokagio:
```suggestion
        // Always provide a value in order to determine how often we're unable to application state
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
 
-        shouldSendEventCallback?(event, shouldSendEvent)
-
-        guard let eventLogging = self.eventLogging else {
-            DDLogDebug("📜 Cancelling log file attachment – Event Logging is not initialized")
-            return shouldSendEvent
+        /// If we shouldn't send the event we have nothing else to do here
+        guard shouldSendEvent else {
+            return nil
         }
```
mokagio: What do you think of moving this check at the very start of the method? If we don't need to send the event, then there's no reason to post-process the tags.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
-        let shouldSendEvent = !CrashLogging.userHasOptedOut
+        let shouldSendEvent = !dataProvider.userHasOptedOut
         #endif
 
-        shouldSendEventCallback?(event, shouldSendEvent)
-
-        guard let eventLogging = self.eventLogging else {
-            DDLogDebug("📜 Cancelling log file attachment – Event Logging is not initialized")
-            return shouldSendEvent
+        /// If we shouldn't send the event we have nothing else to do here
```
mokagio, line 64:
```suggestion
        // If we shouldn't send the event we have nothing else to do here
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+        guard shouldSendEvent else {
+            return nil
         }
 
-        guard let event = event else {
-            DDLogDebug("📜 Cancelling log file attachment – event is nil")
-            return shouldSendEvent
-        }
+        /// Read the current user from the Data Provider (though the Data Provider can decide not to provide it for functional or privacy reasons)
```
mokagio, line 78:
```suggestion
        // Read the current user from the Data Provider (though the Data Provider can decide not to provide it for functional or privacy reasons)
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+        event.user = dataProvider.currentUser?.sentryUser
 
-        guard shouldSendEvent else {
-            DDLogDebug("📜 Cancelling log file attachment – should not send event")
-            return shouldSendEvent
+        /// Everything below this line is related to event logging, so if it's not set up we can exit
```
mokagio, line 81:
```suggestion
        // Everything below this line is related to event logging, so if it's not set up we can exit
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
 
-    /**
-     Writes the error to the Crash Logging system, and includes a stack trace.
+    ///
+    /// Writes the error to the Crash Logging system, and includes a stack trace.
+    /// - Parameters:
+    ///   - error: The error object
+    ///   - userInfo: A dictionary containing additional data about this error.
+    ///   - level: The level of severity to report in Sentry (`.error` by default)
+    func logError(_ error: Error, userInfo: [String: Any]? = nil, level: SentryLevel = .error) {
```
mokagio, line 112: _Future_ refinement idea. `CrashLogging` is cool because it hides Sentry from the consumers. 

The abstraction it's somehow leaky because an app using it will still need to know about Sentry to an extent to pass the API key and what not. Still, from an API point of view, everything is behind `CrashLogging`, which is neat.

We could push the abstraction even further and define a `LogLevel` enum in `CrashLogging` to use as the input for this method.

If you think this is a good suggestion, I'll hope an issue or a PR directly.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
-     - error: The error object
-     - userInfo: A dictionary containing additional data about this error.
-     - level: The level of severity to report in Sentry (`.error` by default)
-    */
-    static func logError(_ error: Error, userInfo: [String: Any]? = nil, level: SentrySeverity = .error) {
-        let event = Event(level: .error)
-        event.message = error.localizedDescription
+        let event = Event(level: level)
+
+        /// Use the unlocalized error message for better grouping
```
mokagio, line 116:
```suggestion
        // Use the unlocalized error message for better grouping
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+        event.message = SentryMessage(formatted: (error as NSError).description)
+
+        /// If the developer provides their own userInfo, use that – otherwise read it from the Error
```
mokagio, line 119:
```suggestion
        // If the developer provides their own userInfo, use that – otherwise read it from the Error
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
         event.extra = userInfo ?? (error as NSError).userInfo
+
+        /// Attach the localized description in case it has additional data
```
mokagio, line 122:
```suggestion
        // Attach the localized description in case it has additional data
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
 }
 
 // Manual Error Logging
 public extension CrashLogging {
 
-    /**
-     Writes the error to the Crash Logging system, and includes a stack trace.
+    ///
+    /// Writes the error to the Crash Logging system, and includes a stack trace.
+    /// - Parameters:
```
mokagio:
```suggestion
    /// Writes the error to the Crash Logging system, and includes a stack trace.
    ///
    /// - Parameters:
```

Other headers in this file have a neat new line to separate description for parameters. This suggestion makes this consistent with them.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+
+    func logErrorsImmediately(_ errors: [Error], userInfo: [String: Any]? = nil, level: SentryLevel = .error, callback: @escaping (Result<Bool, Error>) -> Void) throws {
+
+        var serializer = SentryEventSerializer(dsn: dataProvider.sentryDSN)
+
+        errors.forEach {
+            let event = Event(level: level)
+            event.message = SentryMessage(formatted: $0.localizedDescription)
+            event.timestamp = Date()
+            event.extra = userInfo ?? ($0 as NSError).userInfo
```
mokagio:
```suggestion
        errors.forEach { error in
            let event = Event(level: level)
            event.message = SentryMessage(formatted: error.localizedDescription)
            event.timestamp = Date()
            event.extra = userInfo ?? (error as NSError).userInfo
```

Nitpick: I was surprised by the `$0` because I've always seen it in one liner closures. It might be just me...

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+        guard let authString = dsn.getAuthString() else {
+            throw Errors.unableToConstructAuthStringError
+        }
+
+        var request = URLRequest(url: dsn.getEnvelopeEndpoint())
+        request.httpMethod = "POST"
+        request.httpBody = requestBody
+        request.addValue(authString, forHTTPHeaderField: "X-Sentry-Auth")
+
+        URLSession.shared.dataTask(with: request) { (responseData, urlResponse, error) in
```
mokagio, line 185: Future refinement suggestion: what do you think of injecting a `URLSession` instance at `init` time (maybe with default `.shared` value) to avoid this direct singleton access?

It could eventually help us to test the callback calling behavior.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+        semaphore.wait()
+
+        if let networkError = networkError {
+            throw networkError
+        }
+    }
 
-        Client.shared?.snapshotStacktrace {
-            Client.shared?.appendStacktrace(to: event)
+    /// A wrapper around the `SentryClient` shim – keeps each layer clean by avoiding optionality
```
mokagio, line 228: 👍

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -2,178 +2,251 @@ import Foundation
...
+    private func tryAddingStackTrace(to event: Event) -> Event {
```
mokagio: Nitpick. I found it a bit odd to read "try" as part of the method name.

Having said that, I don't have a solid alternative to suggest. 🤔 

Maybe `addStackTraceWhenPossible(to event:)`?

Automattic-Tracks-iOS/Crash Logging/SentryClient+Shim.m
```diff
@@ -0,0 +1,39 @@
...
+                          isCrashEvent:(BOOL)isCrashEvent;
+@end
+
+@implementation SentryClient(Shim)
+
+/*
+ * A wrapper around a private method in `SentryClient`. If it's unavailable (say the library changes),
+ * the worst that should happen is that the stack trace is no longer available on events sent manually using this method.
+ *
+ * We can remove this once the Sentry SDK allows for capturing events and being notified once they're delivered.
```
mokagio, line 16: Is there a GitHub issue on their end we can link here?

Automattic-Tracks-iOS/Crash Logging/SentryEventSerializer.swift
```diff
@@ -0,0 +1,58 @@
...
+        self.jsonEncoder = JSONEncoder()
+        self.jsonEncoder.dateEncodingStrategy = .iso8601
+    }
+
+    mutating func add(event: Event, filename: String = UUID().uuidString) {
+        events[filename] = event
+    }
+
+    func serialize(eventId: UUID = UUID()) throws -> Data {
+        /// Turn the UUID into a `SentryId` temporarily, just in case they ever change the API on us – that way it won't fail
```
mokagio, line 22:
```suggestion
        // Turn the UUID into a `SentryId` temporarily, just in case they ever change the API on us – that way it won't fail
```

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -0,0 +1,97 @@
+import SwiftUI
```
mokagio, line 1: 😱

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -0,0 +1,97 @@
...
+import Sentry
+
+@available(iOS 13.0, OSX 10.15, *)
+public struct CrashLoggingView: View {
+
+    @State
+    var sendErrorAndWaitStatus: SendErrorAndWaitStatus = .none
```
mokagio, line 8: Genuinely curious: I've never seen property wrappers being defined on the line above their property. Where does this style come from?

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -0,0 +1,97 @@
...
+            Section(header: Text("Actions")) {
+                Button("Send Test Crash", action: sendTestCrash)
+                Button("Send Test Event", action: sendTestEvent)
+                HStack {
+                    Button(action: sendErrorAndWait) {
+                        HStack {
+                            Text("Send Error and Wait")
+
+                            /// Nicely align the icon to the right on iOS but don't leave
+                            /// a weird-shaped button on macOS
```
mokagio, line 30:
```suggestion
                            // Nicely align the icon to the right on iOS but don't leave
                            // a weird-shaped button on macOS
```

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -0,0 +1,97 @@
...
+                            #if os(iOS)
+                            Spacer()
+                            #endif
+
+                            switch sendErrorAndWaitStatus {
+                                case .none:
+                                    Group {} /// An empty view
```
mokagio, line 37:
```suggestion
                                    Group {} // An empty view
```

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -0,0 +1,97 @@
...
+                                    Text("✅")
+                                case .error:
+                                    Text("⚠️")
+                            }
+                        }
+                    }
+                }
+            }
+
+            /// Push the form to the top of the screen
```
mokagio, line 50:
```suggestion
            // Push the form to the top of the screen
```

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -0,0 +1,97 @@
...
+                            switch sendErrorAndWaitStatus {
+                                case .none:
+                                    Group {} /// An empty view
+                                case .uploading:
+                                    Text("⏳")
+                                case .success:
+                                    Text("✅")
+                                case .error:
+                                    Text("⚠️")
+                            }
```
mokagio, line 44: I'm a big fan of using enums to describe mutually exclusive states in the UI. Nicely done 😄

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -0,0 +1,97 @@
...
+        self.crashLogging = crashLogging
+    }
+
+    public var body: some View {
+        Form {
+            Section(header: Text("Actions")) {
+                Button("Send Test Crash", action: sendTestCrash)
+                Button("Send Test Event", action: sendTestEvent)
+                HStack {
+                    Button(action: sendErrorAndWait) {
```
mokagio, line 25: I'm obviously over thinking this, given this is a debug view, but... as far as I can see, there is no logic to prevent running the action when the state is `.uploading`.

Automattic-Tracks-iOSTests/Tests/CrashLoggingTests.swift
```diff
@@ -9,176 +9,76 @@ class CrashLoggingTests: XCTestCase {
...
-    func testThatUserDataIsBeingStoredForUseInCrashLogs() {
+    func testThatUserDataIsBeingPopulatedWhenSendingCrashLogs() throws {
         mockDataProvider.currentUser = testUser
-        CrashLogging.start(withDataProvider: mockDataProvider)
+        let crashLogging = try CrashLogging(dataProvider: mockDataProvider).start()
+        let event = crashLogging.beforeSend(event: Event(level: .debug))
 
-        XCTAssert(CrashLogging.sharedInstance.currentUser.email == testUser.email)
-        XCTAssert(CrashLogging.sharedInstance.cachedUser?.email == testUser.email)
+        XCTAssertEqual(testUser.email, event?.user?.email)
```
mokagio, line 37: Nitpick. It's unclear in the signature, but the convention I've always seen used with `XCTAssertEqual` it to have the actual value as the first element and the expect as the second.

That's what the Apple examples do, too:

Screen Shot 2021-01-20 at 12 31 00 pm

[source](https://developer.apple.com/documentation/xctest/2142776-xctassertequal)

image

[source](https://developer.apple.com/documentation/xctest/defining_test_cases_and_test_methods).

Is there a different convention in place in this codebase?

TracksDemo/Podfile
```diff
@@ -4,16 +4,16 @@ use_frameworks!
 target 'TracksDemo' do
   platform :ios, '12.0'
 
-  pod 'Automattic-Tracks-iOS', :path => '../'
+  pod 'Automattic-Tracks-iOS', path: '../'
```
mokagio, line 7: 🥳 _it's the little things..._

mokagio (commented):
I need to take a break, so I'm submitting this incomplete review.

So far everything looks good. I really only left nitpicks.

TracksDemo/Shared/CrashLoggingDataSource.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
```
mokagio:
```suggestion
```

Nitpick: Foundation shouldn't be required to build this file.

TracksDemo/Shared/Secrets.example.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
```
mokagio:
```suggestion 
```

Nitpick: Foundation shouldn't be required to build this file.

TracksDemo/TracksDemo/TracksDemo-Bridging-Header.h
```diff
@@ -0,0 +1,4 @@
+//
+//  Use this file to import your target's public headers that you would like to expose to Swift.
+//
+
```
mokagio, line 4: Are we ever going to need a bridging header in this demo project? If not, should we remove this?

TracksDemo/TracksDemo.xcodeproj/project.pbxproj
```diff
@@ -708,14 +765,18 @@
 			baseConfigurationReference = 0262514F2B990CA2F3D99C07 /* Pods-TracksDemo.debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				GCC_PREFIX_HEADER = TracksDemo/TracksDemo_Prefix.pch;
 				INFOPLIST_FILE = TracksDemo/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 14.0;
```
mokagio: The `Podfile` for the demo project sets 12.0 as the deployment target. We can't use that here because of SwiftUI, but should we make them meet in the middle at 13.0?

I can see an argument for running on the latest iOS only and not have to worry about backward compatibility. At the same time, Tracks is not restricted to iOS 14.0 only, so it might be beneficial for the demo app to run against as many compatible versions of iOS as reasonable - that is, iOS 13 so we can keep test the SwiftUI-related stuff.

TracksDemo/TracksDemo.xcodeproj/project.pbxproj
```diff
@@ -792,7 +858,7 @@
 				GCC_C_LANGUAGE_STANDARD = gnu11;
 				INFOPLIST_FILE = "TracksDemo Mac/Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks";
-				MACOSX_DEPLOYMENT_TARGET = 10.14;
+				MACOSX_DEPLOYMENT_TARGET = 11.0;
```
mokagio: Similar comment as the iOS 13.0 vs 14.0 above. Should we set this as 10.15 and do the same on the `Podfile`?

TracksDemo/TracksDemo/CrashLoggingRootViewController.swift
```diff
@@ -0,0 +1,21 @@
+import SwiftUI
+import AutomatticTracks
+
+class CrashLoggingRootViewController: UIHostingController<CrashLoggingView> {
+
+    let crashLogging = CrashLogging(dataProvider: CrashLoggingDataSource())
+
+    init() {
+        crashLogging.start()
```
mokagio: I had to update this to a brute for the iOS demo to build.

```suggestion
        _ = try! crashLogging.start()
```

TracksDemo/TracksDemo/CrashLoggingRootViewController.swift
```diff
@@ -0,0 +1,21 @@
...
+
+        let rootView = CrashLoggingView(crashLogging: crashLogging)
+        super.init(rootView: rootView)
+    }
+
+    @objc required dynamic init?(coder aDecoder: NSCoder) {
+        crashLogging.start()
```
mokagio: Same as above:

```suggestion
        _ = try! crashLogging.start()
```

mokagio (dismissed):
Requesting changes only because of the demo project not building for iOS, which is an easy fix.

I tested on iOS and Mac with my puny Intel and it worked as described; I saw the events with my data on Sentry. _I used my work email if you want to double check_. 👏 👏 

Love that we have a test application. There's a lot we can do to unit test the heck out of this library, but the Sentry integration will always need a bit of manual testing.

On tiny note: I couldn't find the `iOS/tracks-demo` folder mentioned in the description in the secrets store.

I left _a lot_ of nitpicks, which you should take as what they are: just nitpicks.

In particular, I worry my many `///` vs `//` suggestions will come across as overly pedantic. After adding the first couple, I felt the sunk cost pressure and kept on adding them 😳 I put them there to hopefully help with the conversion, but that's not a requirement for this PR to be merged.

TracksDemo/TracksDemo/CrashLoggingRootViewController.swift
jkmassel: Shoot, sorry about that – fixed in b3ba4d2c5358d4ff91524025db5026d21d37f695 (in a slightly different way)

TracksDemo/TracksDemo/CrashLoggingRootViewController.swift
jkmassel: Same as above

TracksDemo/TracksDemo/TracksDemo-Bridging-Header.h
jkmassel: I figured as long as the app is written in Obj-C we might as well keep it around. If we refactor it to Swift it can go then I was thinking 🤷

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: Yeah, I like the internals thing a lot – done in 389e8c17c6adb2af693c522bdb3352bba0ef2278

jkmassel: Regarding the `///` vs `//` thing – WooCommerce started the trend of using the `///` format for comments because they support markdown, links, etc (and are far smaller).

There's some of that already in the ExPlat stuff, so I've been aiming to use the `///` format everywhere. However, I did have instances of `//` that _should_ have been annotated with `MARK:`, so I've added that where appropriate.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: The `fixme` might not be right there? Not sure – I adjusted it in 44f20320584702d5d836cdab6fd26f3fbf271ceb to what I think it should be 😅

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
AliSoftware: Personally I tend to use that style a lot. I think that's even part of most [style guides](https://google.github.io/swift/#attributes) nowadays.

My 2¢ is that this helps in multiple situations:
 - It clearly separates the PW from the rest, making the declaration easier to read, especially in case you end up with multiple annotations in the future
 - It makes for simpler and nicer git diffs if/when you add/remove more PWs on the method
 - It allows to add parameters to the PW annotations without ending up with super long lines because of the longer attributes and parameters
 - It also has the benefit of being consistent with the style used for annotations in Android

Like imagine having all the attributes on one line in this example below…
```swift
@objc(objCFoo)
@available(*, unavailable, renamed: "bar")
@discardableResult
func foo(someParameter: Int, anotherParameter: String) {
  …
}
```

Granted, for properties that have a single and short attribute annotation like just `@State var` I still see a lot of cases where people put it on the same line "because it's short enough". But I prefer the rule to be consistent (and always put them on their own line) rather that changing the style depending on a "feeling" of if it's short enough or not and not be scalable for future changes 😉

AliSoftware:
## Podfile.lock changes / Sentry 6.1.3

While trying to test this, `bundle exec pod install` failed to resolve the dependency for Sentry:
```
$ be pod install
Analyzing dependencies
[!] CocoaPods could not find compatible versions for pod "Sentry":
  In snapshot (Podfile.lock):
    Sentry (= 6.1.2, ~> 6)

  In Podfile:
    Automattic-Tracks-iOS (from `../`) was resolved to 0.7.0, which depends on
      Sentry (~> 6)


You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * changed the constraints of dependency `Sentry` inside your development pod `Automattic-Tracks-iOS`.
   You should run `pod update Sentry` to apply changes you've made.
```
I had to `bundle exec pod update Sentry` first, which led to a bump from Sentry 6.1.2 to 6.1.3 and left the `Podfile.lock` with a diff accordingly.

## Testing and the Secrets.swift file

Also, regarding the `Shared/Secrets.swift`, for anyone who might have been as confused as I was (cc @mokagio) trying to follow the testing instructions in PR description:
 * The credentials to use in the `Shared/Secrets.swift` file are not in our _Secret Store_ in MC… but in our _Mobile Secrets_ repo instead.
 * Which also means there's no need to `cp Shared/Secrets.example.swift Shared/Secrets.swift` in the first place (like I did and guessed I needed to do before trying to find the DSN in Secret Store) because the `Secrets.swft` file is already present in full in our Mobile Secrets repo (i.e. not just the sentryDSN to use for tests, but the whole swift file), so that's the one we should copy for testing.

[EDIT: I took the liberty to edit the PR description to take those remarks ☝️  into account]

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
AliSoftware: ❤️

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -5,188 +5,259 @@ import CocoaLumberjack
...
     }
 
-    /**
-     Initializes the crash logging system.
+    private let dataProvider: CrashLoggingDataProvider
+    private let eventLogging: EventLogging?
 
-     - Parameters:
-     - dataProvider: An object that will provide any required data to the crash logging system.
+    public init(dataProvider: CrashLoggingDataProvider, eventLogging: EventLogging? = nil) {
```
AliSoftware, line 21: Would be nice to add some `///` doc comments here – as well as on the `CrashLoggingDataProvider` and `EventLogging` protocols themselves in the other respective files – to roughly explain what's expected for each of those params

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -5,188 +5,259 @@ import CocoaLumberjack
...
-        isStarted = true
+        /// Validate the DSN ourselves before initializing, because the SentrySDK silently prints the error to the log instead of telling us if the DSN is valid
+        _ = try SentryDsn(string: self.dataProvider.sentryDSN)
 
-        // Store the data provider and event logging subsystem for future use
-        sharedInstance.dataProvider = dataProvider
-        sharedInstance.eventLogging = eventLogging
+        SentrySDK.start { options in
+            options.dsn = self.dataProvider.sentryDSN
+            options.debug = true // Helpful to see what's going on
```
AliSoftware: This comment has some captain obvious vibes, not sure it's worth it? ^^

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -5,188 +5,259 @@ import CocoaLumberjack
...
-     - userInfo: A dictionary containing additional data about this error.
-     - level: The level of severity to report in Sentry (`.error` by default)
-    */
-    static func logError(_ error: Error, userInfo: [String: Any]? = nil, level: SentrySeverity = .error) {
-        let event = Event(level: .error)
-        event.message = error.localizedDescription
+        let event = Event(level: level)
+
+        /// Use the unlocalized error message for better grouping
+        event.message = SentryMessage(formatted: (error as NSError).description)
```
AliSoftware, line 117: Is there a risk that the `Error` (Swift error) we're trying to load and pass to this method is NOT an `NSError` and that this could crash at runtime?

I know that the `as` (not `as!` nor `as?`) here seems to imply that the compiler seems happy with it, but I've seen cases where even if `Swift.Error` was toll-free-bridgeable from ObjC's `NSError`, if the original Swift Error was not created from an `NSError` in the first place it might not have all its fields set…

For example what happens if we have `enum MyError: Error { case foo(code: Int) }` and then call `logError(MyError.foo(code: 42), …)`? What would the `(error as NSError).description` return here?

---

EDIT: I tested it in a Playground and it seems in those cases the description is derived from the debug description, e.g. `__lldb_expr_3.MyError.foo(code: 42)`, but playgrounds are running in debug mode so wonder if that's still true in release.

AliSoftware (changes requested):
After testing the demo project on iOS and macOS on an Intel mac (no M1 here :/), this doesn't seem to be working for me?

 - I only see one new event in Sentry, not two. I think the one I get is from "Send Error and Wait" though not entirely sure because it might just be because of processing delay from Sentry, and also because it's hard to identify those events, given…
 - The only new event I see in Sentry is titled "`The operation couldn’t be completed. (AutomatticTracks.SentryTestError error 1.)`".
    - This matches the `SentryTestError` type indeed
    - But according to what I read in the code, I would have expected an error code of `-1` not `1` (`let error = SentryTestError(title: "Test Error")` with `SentryTestError.init(title: String, code: Int = -1)`)?
    - Also the title displayed in Sentry ("`The operation couldn’t be completed.`") doesn't seem to correspond to the event we expect to have triggered?
    - Finally, maybe it's just me not knowing how to navigate properly in Sentry, but I can't manage to find the `"Test Event <some UUID>"` text (parameter we pass to `crashLogging.logMessage`) nor the `"Test Error"` text (title from the `SentryTestError` we send) anywhere in that Sentry entry/event (I expected them to be in the "message" at the top of the Sentry entry, instead of the "The operation couldn't be completed." message

In any case, doesn't seem to be fully working as expected to me  😢. Even if we solve the message and code mismatch, the fact that I seem to only get one new event instead of two seems unexpected.

Automattic-Tracks-iOS/Crash Logging/SentryTestError.swift
```diff
@@ -0,0 +1,11 @@
...
+
+public struct SentryTestError: LocalizedError {
+    public let title: String
+    public let code: Int
+
+    public init(title: String, code: Int = -1) {
+        self.title = title
+        self.code = code
+    }
+}
```
AliSoftware, line 16: After some investigation and playing with the code, the String `The operation couldn't be completed.` comes from when an `error.localizedDescription` of any `Swift.Error` doesn't provide a custom implementation.
Which is the case here, and explains why we don't see the details of the error title nor code in the Sentry event.

I've modified the code locally to implement one, and it indeed solved it and finally made the `title` and `code` of the error finally surface and be visible in Sentry.

```suggestion
    public var errorDescription: String? {
        "\(title) (code \(code))"
    }
}
```

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -5,188 +5,259 @@ import CocoaLumberjack
...
+        try logErrorsImmediately([error], userInfo: userInfo, level: level, callback: callback)
+    }
+
+    func logErrorsImmediately(_ errors: [Error], userInfo: [String: Any]? = nil, level: SentryLevel = .error, callback: @escaping (Result<Bool, Error>) -> Void) throws {
+
+        var serializer = SentryEventSerializer(dsn: dataProvider.sentryDSN)
+
+        errors.forEach {
+            let event = Event(level: level)
+            event.message = SentryMessage(formatted: $0.localizedDescription)
```
AliSoftware: Some additional insight from Playground testing:

```swift
import Foundation

enum MyError: LocalizedError {
    case foo(code: Int)
}

let e = MyError.foo(code: 42)

print((e as NSError).description) // __lldb_expr_9.MyError.foo(code: 42)
print(String(describing: e)) // foo(code: 42)
print(e.localizedDescription) // The operation couldn’t be completed. (__lldb_expr_9.MyError error 0.)
```

Notice how the last one is what we end up seeing in Sentry's dashboard, because the `SentryTestError` we use in our example project doesn't have a custom `errorDescription: String?` implementation, despite being declared as `LocalizedError` (the default implementation for `LocalizedError.errorDescription` being what you see in the last line of my playground sample above).

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
mokagio: Okay. Maybe I'm wrong about Apple's desired purpose for `///` vs `//`.

Anyways, it's just a style thing 👌

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
mokagio: 🤦 had to look at it for a good 30 seconds to realize what was going wrong 😳 .

44f2032 looks good 👍

TracksDemo/TracksDemo/TracksDemo-Bridging-Header.h
mokagio: 👍

TracksDemo/TracksDemo/CrashLoggingRootViewController.swift
mokagio: Slightly different but still effective 👍

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: I like this idea a lot – this is done in 389e8c17c6adb2af693c522bdb3352bba0ef2278

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: > Maybe I'm wrong about Apple's desired purpose

Probably not – we just abuse it for other things 😅

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: Yeah good call – done in b0eedd7f2a41ab95ed3eef18051977078afd4ddb

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: Yeah good call – I like the idea of hiding that implementation detail. 

In general, given the Sentry uses a subset of the RFC 5424 severity levels I'd imagine that if someone wrote `.error`, it'd be a non-breaking change, but yeah, might as well wrap it

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: Works for me – changed in baec210c6d4af873da6fe2de90e6e15ebd727208

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: Yep, agreed. Eventually I'd like for all of the `Tracks` functionality to be able to use an injected session that can be defined by the hosting app – would make sense to roll out to all the things for sure.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: I can see where you're coming from here – I just went with `addStackTrace(to event:)`. 

For the most part this shouldn't fail – the main failure mode would be if the API changed out from underneath us. 

Updated in 20af629f562ad08ff3c937a810a237b868e95a72

Automattic-Tracks-iOS/Crash Logging/SentryClient+Shim.m
jkmassel: Yeah, for sure – added in c6a751b926031609d8fe5eb49cbc2aac85589ddc

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
jkmassel: I just started doing it because some variable names got quite long and it was a pain in the neck to read. If I was going to be consistent I fell down on the side of separate lines, but @AliSoftware's rationale is way better than my "whoops I fell into this" approach 😄

Automattic-Tracks-iOS/UI/Crash Logging/CrashLoggingView.swift
jkmassel: Agreed. I had to strongly resist the urge to over-build this because it _is_ a demo app, so I'll stay strong and leave this for now.

Though it will bug me lol

Automattic-Tracks-iOSTests/Tests/CrashLoggingTests.swift
jkmassel: No – that convention is correct!

Here, the `testUser` is the "constant" (ie – not the subject of the test) so the ordering is as expected.

Definitely something to nitpick – IMHO this is important for test readability so we should keep it consistent!!

TracksDemo/TracksDemo.xcodeproj/project.pbxproj
jkmassel: Yeah seems reasonable – changed to iOS 13 in 9ed5b7a08e4ec26b97358489e1741d2cd5451c7c

TracksDemo/TracksDemo.xcodeproj/project.pbxproj
jkmassel: Addressed in 9ed5b7a08e4ec26b97358489e1741d2cd5451c7c, Podfile in 58f90120b5e589774357aacb8fe353e48aa72663

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: Yeah, I agree! Added in 6d6c33f585d5ff8c3c5cf4dda7f89f41809ab310

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
jkmassel: LOL yes – removed in a571403866b932ed157f3b0f0d08c2efc65982aa

Automattic-Tracks-iOS/Crash Logging/SentryTestError.swift
jkmassel: Ah sorry, yeah as you now know we were hitting that path with every test error – I just happened to know what to look for in Sentry 😅

I like this for its obviousness in these situations.

Addressed in 9b9c107b2cacca1cd986d31619366a36a341a414 – I just added a bit of a description to the `errorDescription` field

jkmassel:
@mokagio and @AliSoftware, thanks for the feedback – it should all be addressed now – ready for another look when you're ready 😃 

One last change – I didn't realize the schemes weren't shared – I've added them to the repo in 901bf1d3a3aa82ef4201e223892323efd400db91. That should fix the issue where in testing the "Send Test Event" wouldn't work (it needs to override the `debug` mode).

AliSoftware:
Tested on iPhone Simulator (iPhone 11 Pro, iOS 13.5) and on Intel Mac. Was able to see both events in Sentry ("Send Test Event" + "Send Error and Wait") in both cases

I got this log in Xcode console each time I clicked the button when testing with the iPhone Simulator app. I'm not entirely sure it's expected to especially get the "Cancelling log file attachment"? 🤔 

```
2021-01-22 16:25:22:594 TracksDemo[77872:5993852] 📜 Firing `beforeSend`
2021-01-22 16:25:22.594496+0100 TracksDemo[77872:5994015] 📜 Firing `beforeSend`
2021-01-22 16:25:22:594 TracksDemo[77872:5993852] 📜 This is a debug build
2021-01-22 16:25:22.594710+0100 TracksDemo[77872:5994015] 📜 This is a debug build
2021-01-22 16:25:22:594 TracksDemo[77872:5993852] 📜 Cancelling log file attachment – Event Logging is not initialized
2021-01-22 16:25:22.594901+0100 TracksDemo[77872:5994015] 📜 Cancelling log file attachment – Event Logging is not initialized
```


I also _still_ got a diff on the `Podfile.lock` even after cleaning my `Pods/` and working copy and redoing a `pod install`… but this time a different one.

<details><summary>git clean + pod install + git diff</summary>

```
$ git clean -dxf
Removing Pods/
Removing Shared/Secrets.swift
Removing TracksDemo.xcodeproj/xcuserdata/
Removing TracksDemo.xcworkspace/xcuserdata/
                                                                                                                                                                                                                                
$ cp ~/.mobile-secrets/iOS/tracks-demo/Secrets.swift Shared/Secrets.swift         
                                                                                                                                                                                                                                
$ be pod install
Analyzing dependencies
Downloading dependencies
Installing Automattic-Tracks-iOS (0.7.1)
Installing CocoaLumberjack (3.4.2)
Installing Reachability (3.2)
Installing Sentry (6.1.2)
Installing Sodium (0.9.1)
Installing UIDeviceIdentifier (1.1.4)
Generating Pods project
Integrating client project
Pod installation complete! There is 1 dependency from the Podfile and 6 total pods installed.
                                                                                                                                                                                                                                
$ gst
On branch update/sentry-to-v6
Your branch is up to date with 'origin/update/sentry-to-v6'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Podfile.lock

no changes added to commit (use "git add" and/or "git commit -a")

$ git diff

diff --git a/TracksDemo/Podfile.lock b/TracksDemo/Podfile.lock
index 712c51e..f701edb 100644
--- a/TracksDemo/Podfile.lock
+++ b/TracksDemo/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Automattic-Tracks-iOS (0.7.0):
+  - Automattic-Tracks-iOS (0.7.1):
     - CocoaLumberjack (~> 3)
     - Reachability (~> 3)
     - Sentry (~> 6)
@@ -34,7 +34,7 @@ EXTERNAL SOURCES:
     :path: "../"
 
 SPEC CHECKSUMS:
-  Automattic-Tracks-iOS: 01fd960098c3e6e2822531c1eeb6d03dbddd865c
+  Automattic-Tracks-iOS: 7ac2cc974ec22d1273bd8df9eae0cfb3193d959d
   CocoaLumberjack: db7cc9e464771f12054c22ff6947c5a58d43a0fd
   Reachability: 33e18b67625424e47b6cde6d202dce689ad7af96
   Sentry: 3a6c3aeb258b297a4be08d8960955ecc722331f3
```

</details>

Automattic-Tracks-iOSTests/Tests/CrashLoggingTests.swift
AliSoftware: Is it though? @mokagio 's suggestion is that the convention is `XCTAssert(sut, expectedConstantValue)`, so the constant being as the second parameter. That's also always the convention I used I think.
So if the SUT is `event?.user?.email` and the `testUser` is the "constant" (= the expected value), then the order in your code is reversed, isn't it?

jkmassel:
> I got this log in Xcode console each time I clicked the button when testing with the iPhone Simulator app. I'm not entirely sure it's expected to especially get the "Cancelling log file attachment"? 🤔

That's expected – the encrypted logging system isn't injected into `CrashLogging`, so we log the fact that we're not uploading an encrypted log for this crash. This is especially important when debugging in-app so devs aren't confused why no logs are showing up and there's no `logId` associated with the Sentry event.

jkmassel:
Thanks for the detailed instructions – I was able to replicate that issue this time!

This time I think it was caused by the merge with `develop`? 

But should be resolved now 😅

Automattic-Tracks-iOS/Crash Logging/SentryClient+Shim.m
```diff
@@ -0,0 +1,40 @@
...
+    SentryScope *scope = [[SentrySDK currentHub] getScope];
+    SentryEvent *eventWithStackTrace = [self prepareEvent:event withScope:scope alwaysAttachStacktrace:YES isCrashEvent:NO];
+
+    if(eventWithStackTrace == nil) {
+        return event;
+    }
+
+    if(eventWithStackTrace.threads.firstObject.stacktrace != nil) {
+        eventWithStackTrace.stacktrace = eventWithStackTrace.threads.firstObject.stacktrace;
+    }
```
AliSoftware: I am guessing this is based on the assumption that the real crash is _always_ guaranteed to be in the `threads.firstObject`? I mean I guess that makes sense given that that's probably the thread that was active when the crash occurred, but is the currently active thread always the one that is first in this array though? 🤔

Automattic-Tracks-iOS/Crash Logging/SentryEventSerializer.swift
```diff
@@ -0,0 +1,58 @@
...
+        let header = SentryHeader(event_id: SentryId.init(uuid: eventId).sentryIdString, dsn: dsn)
```
AliSoftware:
```suggestion
        let header = SentryHeader(event_id: SentryId(uuid: eventId).sentryIdString, dsn: dsn)
```

Automattic-Tracks-iOS/Crash Logging/SentryClient+Shim.m
```diff
@@ -0,0 +1,40 @@
...
+ * the worst that should happen is that the stack trace is no longer available on events sent manually using this method.
+ *
+ * We can remove this once the Sentry SDK allows for capturing events and being notified once they're delivered.
+ * ref: https://github.com/getsentry/sentry-cocoa/issues/660
+ */
+-(SentryEvent *) addStackTraceToEvent:(SentryEvent *) event forClient:(SentryClient *) client {
+
+    SEL selector = @selector(prepareEvent:withScope:alwaysAttachStacktrace:isCrashEvent:);
+
+    if(![client respondsToSelector:selector]) {
```
AliSoftware: nit:
```suggestion
    if (![client respondsToSelector:selector]) {
```

Automattic-Tracks-iOS/Crash Logging/SentryClient+Shim.m
```diff
@@ -0,0 +1,40 @@
...
+        return event;
+    }
+
+    SentryScope *scope = [[SentrySDK currentHub] getScope];
+    SentryEvent *eventWithStackTrace = [self prepareEvent:event withScope:scope alwaysAttachStacktrace:YES isCrashEvent:NO];
+
+    if(eventWithStackTrace == nil) {
```
AliSoftware: nit:
```suggestion
    if (eventWithStackTrace == nil) {
```

Automattic-Tracks-iOS/Crash Logging/SentryClient+Shim.m
```diff
@@ -0,0 +1,40 @@
...
+        return event;
+    }
+
+    if(eventWithStackTrace.threads.firstObject.stacktrace != nil) {
```
AliSoftware: nit:
```suggestion
    if (eventWithStackTrace.threads.firstObject.stacktrace != nil) {
```

AliSoftware (approved):
Took some time to also re-read the code.

Overall LGTM; I left a couple of nits but that are optional and shouldn't block that PR.
The other nits about `XCTAssertEqual(sut, expected)` as convention also still apply (but are also only nits, so…)

I still left one question about the assumption for active thread that might need checking and validating before merging though (but didn't want to keep the "Request Changes" status just for this one.) so I'll let you address them before hitting the button.

Automattic-Tracks-iOS/Crash Logging/SentryClient+Shim.m
jkmassel: Yeah, for now it's a reasonable assumption – this functionality is designed for logging crashes on startup so that we can log the message before another crash happens. The startup stuff is all currently on the main thread, so this should work fine for quite some time.

As part of the core project to re-work app initialization, I'm hoping to kill this code and be able to put up a UI for something like "the app crashed last time – would you like to try again or reset?" and in that time Sentry could do its thing and we could kill this code.

mokagio: >  put up a UI for something like "the app crashed last time – would you like to try again or reset?"

That would be such a great thing to have!

loremattei (approved):
I haven't re-reviewed the code as I see that @AliSoftware and @mokagio already did it. 

I tested on a M1 Mac following the instructions: 
☑️ Try building for Any iOS Device
☑️ Try building for My Mac, and Any Mac (Apple Silicon, Intel)

and everything (including archiving it) was successful. 
I've seen a few warnings, but I suppose they are expected at this point.

PR #167, opened 2021-01-19
ExPlat: remove "unknown" variation, handle null and not-returned cases

leandroalonso:
This PR implements some changes in the AB testing feature based on @withinboredom comments in a [previous PR](https://github.com/Automattic/Automattic-Tracks-iOS/pull/155#pullrequestreview-571332826):

* Remove the `unknown` variation. The default fallback is `control`
* If the experiment has a `null` value in the JSON, treat it as `control`
* If the experiment is not returned in the JSON, treat it as `control`
* Removes the `other()` variation. All variations are effectively a treatment variation, the API was slightly changed:

```swift
            switch abTesting.experiment("experiment") {
            case .treatment(let variation) where variation == "variation-name":
                // do something
            case .treatment(_):
                // default variation
            case .control:
                // control
            }
```

### How to test

Code review and green CI should be enough. :)

### DIFF
--- a/Automattic-Tracks-iOS/ABTesting/ABTesting.swift
+++ b/Automattic-Tracks-iOS/ABTesting/ABTesting.swift
@@ -4,7 +4,6 @@ public enum Variation: Equatable {
     case control
     case treatment
     case other(String)
-    case unknown
 }
 
 /// A protocol that defines a A/B Testing provider
--- a/Automattic-Tracks-iOS/ABTesting/ABTesting.swift
+++ b/Automattic-Tracks-iOS/ABTesting/ABTesting.swift
@@ -2,8 +2,7 @@ import Foundation
 
 public enum Variation: Equatable {
     case control
-    case treatment
-    case other(String)
+    case treatment(String?)
 }
 
 /// A protocol that defines a A/B Testing provider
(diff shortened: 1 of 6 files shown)


Automattic-Tracks-iOS/ABTesting/ABTesting.swift
```diff
@@ -4,7 +4,6 @@ public enum Variation: Equatable {
     case control
     case treatment
     case other(String)
```
withinboredom: It might be a good idea to do what @renanferrari did on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1845 -- and make Treatment take a string since `other()` is just a different Treatment (for when we support multi-variate experiments).

withinboredom (approved):
LGTM, but might want to also remove `other` as well, and just have `treatment` accept a name for the treatment.

leandroalonso:
@jkmassel I have a totally unrelated Mac OS X test failing here (`testThatOnlyOneFileIsUploadedSimultaneously`).

Running it again doesn't seem to solve (and the test works just fine in my machine™️), any idea how to proceed?

_Edit: of course now it worked._

PR #168, opened 2021-01-26
Cut 0.8.0

jkmassel:
Bump the version number in preparation for a new release

### DIFF
--- a/Automattic-Tracks-iOS/TracksConstants.m
+++ b/Automattic-Tracks-iOS/TracksConstants.m
@@ -1,4 +1,4 @@
 #import "TracksConstants.h"
 
 NSString *const TracksErrorDomain = @"TracksErrorDomain";
-NSString *const TracksLibraryVersion = @"0.7.1";
+NSString *const TracksLibraryVersion = @"0.8.0";
--- a/TracksDemo/Podfile.lock
+++ b/TracksDemo/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Automattic-Tracks-iOS (0.7.1):
+  - Automattic-Tracks-iOS (0.8.0):
     - CocoaLumberjack (~> 3)
     - Reachability (~> 3)
     - Sentry (~> 6)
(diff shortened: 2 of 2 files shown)


Automattic-Tracks-iOS/TracksConstants.m
```diff
@@ -1,4 +1,4 @@
 #import "TracksConstants.h"
 
 NSString *const TracksErrorDomain = @"TracksErrorDomain";
-NSString *const TracksLibraryVersion = @"0.7.1";
+NSString *const TracksLibraryVersion = @"0.8.0";
```
AliSoftware, line 4: Wow was about to say you should update the `podspec` too but I just saw that it actually already [parses that `.m` file dynamically](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Automattic-Tracks-iOS.podspec#L3), nice. Not sure how stable that trick/solution is (e.g. what if we `pod trunk push` while the PWD is not the repo root, not sure the relative path used by `File.read` will work? etc) but something separate to a release cut, so might inspect that separately.

AliSoftware (changes requested):
I think you need to run `pod install` in the demo project too, to [update its `Podfile.lock`](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/TracksDemo/Podfile.lock) – as [the demo project's own `Podfile` references the lib by relative path (development pod)](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/TracksDemo/Podfile#L7) so resolves against the local `podspec` which now (dynamically) defines the version to `0.8.0`.

Automattic-Tracks-iOS/TracksConstants.m
jkmassel: Yeah, I really don't know how I feel about the whole parsing-the-version thing in this repo 😅

AliSoftware (approved):
Looks good with the lockfile fix now 👍

jkmassel:
> I think you need to run pod install in the demo project too, to update its Podfile.lock – as the demo project's own Podfile references the lib by relative path (development pod) so resolves against the local podspec which now (dynamically) defines the version to 0.8.0.

Yeah good call – I'll try to remember to get a linter rule going for this or something, else I'll for sure forget next time

Automattic-Tracks-iOS/TracksConstants.m
AliSoftware: I would not blame you if you decide to remove that parsing part and switch to hardcoded version before merging I you want to 😇

PR #173, opened 2021-03-12
Don’t send localized strings to Sentry

jkmassel:
Currently errors logged using `logError` send localized strings to Sentry, which causes a lot of noise. 

This change will always send the English description instead.

**To Test:**
- Run the demo app, and choose "Send Test Error". The error will show up in Sentry untranslated.

### DIFF
--- a/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
+++ b/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
@@ -10,11 +10,8 @@ extension Event {
     ) -> Event {
         let event = Event(level: level)
 
-        if let underlyingError = error.userInfo[NSUnderlyingErrorKey] as? NSError {
-            event.message = SentryMessage(formatted: underlyingError.debugDescription)
-        } else {
-            event.message = SentryMessage(formatted: error.debugDescription)
-        }
+        let baseError: NSError = (error.userInfo[NSUnderlyingErrorKey] as? NSError) ?? error
+        event.message = SentryMessage(formatted: baseError.debugDescription)
 
         /// The error code/domain can be used to reproduce the original error
         var mutableExtra = extra
--- a/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
+++ b/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
@@ -1,0 +1,33 @@
+import Sentry
+
+extension Event {
+    static func from(
+        error: NSError,
+        level: SentryLevel = .error,
+        user: Sentry.User? = nil,
+        timestamp: Date = Date(),
+        extra: [String: Any] = [:]
+    ) -> Event {
+        let event = Event(level: level)
+
+        if let underlyingError = error.userInfo[NSUnderlyingErrorKey] as? NSError {
+            event.message = SentryMessage(formatted: underlyingError.debugDescription)
+        } else {
+            event.message = SentryMessage(formatted: error.debugDescription)
+        }
+
+        /// The error code/domain can be used to reproduce the original error
+        var mutableExtra = extra
+        mutableExtra["error-code"] = error.code
+        mutableExtra["error-domain"] = error.domain
+
+        error.userInfo.forEach { (key, value) in
+            mutableExtra[key] = value
+        }
+
+        event.timestamp = timestamp
+        event.extra = [String:Any](uniqueKeysWithValues: mutableExtra.sorted { $0.key > $1.key })
+        event.user = user
+        return event
+    }
+}
--- a/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
+++ b/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
@@ -26,7 +26,7 @@ extension Event {
         }
 
         event.timestamp = timestamp
-        event.extra = [String: Any](uniqueKeysWithValues: mutableExtra.sorted { $0.key > $1.key })
+        event.extra = mutableExtra
         event.user = user
         return event
     }
--- a/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
+++ b/Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
@@ -26,7 +26,7 @@ extension Event {
         }
 
         event.timestamp = timestamp
-        event.extra = [String:Any](uniqueKeysWithValues: mutableExtra.sorted { $0.key > $1.key })
+        event.extra = [String: Any](uniqueKeysWithValues: mutableExtra.sorted { $0.key > $1.key })
         event.user = user
         return event
     }
--- a/Automattic-Tracks-iOSTests/Tests/SentryExtensionTests.swift
+++ b/Automattic-Tracks-iOSTests/Tests/SentryExtensionTests.swift
@@ -1,0 +1,23 @@
+import XCTest
+@testable import AutomatticTracks
+import Sentry
+
+class SentryExtensionTests: XCTestCase {
+
+    override func setUpWithError() throws {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
+
+    func testThatEventInitializationReturnsEnglishErrorDescription() {
+        do {
+            _ = try FileManager.default.attributesOfItem(at: URL(fileURLWithPath: "/not-a-real-path"))
+        } catch let err {
+            let event = Event.from(error: err as NSError)
+            XCTAssertEqual(event.message.formatted, "Error Domain=NSPOSIXErrorDomain Code=2 \"No such file or directory\"")
+        }
+    }
+}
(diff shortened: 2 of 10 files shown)


AliSoftware (commented):
Do you think it would be possible to add a unit test for this? 🤔

jkmassel:
> Do you think it would be possible to add a unit test for this? 🤔

Yeah, good question. Possible? Yes. Practical...I'm not sure.

Tracks acts as an abstraction layer over Sentry (or another crash logging tool if we were to switch), so it's kind of definitionally tightly coupled.

That said, we could introduce _another_ abstraction layer between them that allows for testing this kind of thing, but I don't know that that would make it better enough for the additional complexity?

Thats the main reason for the Sentry debugging view – to be able to test this stuff on the rare occasion it needs to change (though I do sort of like the approach to automated testing, I feel like it'd be a lot of "test that the code does what the code does" types of things...)

WDYT?

mokagio (approved):
Tested as described and saw the error message in English.

Screen Shot 2021-03-17 at 2 28 53 pm

I tried it with my Simulator in Italian and I still got the error message in English, though. I think that has to do with the fact the test methods calls `logError` but it was `logErrorsImmediately` that sent the localized descriptions. Plus, it turned out that the Italian localized description of the error was in English anyways 🙃 

---

Regarding the unit test request, an idea could be to separate the logic that builds the event from the logic sending it. We can't easily test the logic that sends the even, and we shouldn't because that's all under Sentry's control. We can test a function or `init` that takes some inputs and returns a configured `Event` in a straightforward way.

```swift
func testErrorEventCreation() {
  let error = ...

  let event = Event(from: error, extra: ["a-key": "a value"])

  XCTAssertEqual(even.level, .error)
  // etc...
}
```

jkmassel:
Good suggestion @mokagio – I've made a static initializer method to convert an `NSError` to an `Event`, and added a test. I've used Test Plans to test in French and English (which was good, because at least in test plans the old method didn't actually work...)

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
```diff
@@ -0,0 +1,33 @@
+import Sentry
+
+extension Event {
+    static func from(
+        error: NSError,
+        level: SentryLevel = .error,
+        user: Sentry.User? = nil,
+        timestamp: Date = Date(),
+        extra: [String: Any] = [:]
+    ) -> Event {
```
AliSoftware, line 10: nit: why not `init(error:level:user:timestamp:extra:)` (or `init(fromError:level:user:timestamp:extra:)`) instead of `static func`? 🤔

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        /// The error code/domain can be used to reproduce the original error
+        var mutableExtra = extra
+        mutableExtra["error-code"] = error.code
+        mutableExtra["error-domain"] = error.domain
+
+        error.userInfo.forEach { (key, value) in
+            mutableExtra[key] = value
+        }
```
AliSoftware, line 23: I'm wondering if that works in all cases – I mean with the fact that `mutableExtra` is `[String: Any]` I wonder how Sentry serialize that extra dictionary when the `Any` values are not JSON-compatible types?

For example, if your `error` has an `userInfo[NSUnderlyingErrorKey]`, whose value is typically itself another `NSError`, that means that `mutableExtra` will end up with one of its value being an `NSError`, and I wonder how that will be serialized and show up in Sentry?

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
```diff
@@ -0,0 +1,33 @@
...
+
+        event.timestamp = timestamp
+        event.extra = [String:Any](uniqueKeysWithValues: mutableExtra.sorted { $0.key > $1.key })
```
AliSoftware: What's the point of sorting the `mutableExtra` by key if you end up putting them in an dictionary, which by definition is a non-ordered collection?

Automattic-Tracks-iOSTests/Tests/SentryExtensionTests.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    override func setUpWithError() throws {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
+
+    func testThatEventInitializationReturnsEnglishErrorDescription() {
```
AliSoftware, line 17:
```suggestion
    // Note: for this test to be useful, it needs to be run as part of a Test Plan which runs the test suite
    // in languages other than English. See TestPlan.xctestplan
    func testThatEventInitializationReturnsEnglishErrorDescription() {
```

Just because, even if a test plan is present, it's possible to run the tests (e.g. via command line or fastlane) without using the test plan explicitly and instead using the default config; and it's not always obvious / very visible that a Test Plan exists in the project and is used, so I think it can be useful to highlight that it does here 😉

AliSoftware (commented):
Some nits and suggestions

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
jkmassel: Yeah, good point – too much time spent with PHP I guess. Though when I added this it started sorting the entries properly in Sentry. But now I removed it and it does too. Gremlins, I guess?

Removed in dfec0bfa7d0b20e5b679d54cbb2ee9c5e24b1b23

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
jkmassel: Sentry doesn't use `Codable` for JSON, so it'll just print the object description for anything that's invalid. Adding 

```swift
        mutableExtra["foo"] = 1 as Any
        mutableExtra["bar"] = DispatchQueue(label: "test")
```

to the list results in this:

Screen Shot 2021-03-24 at 10 34 20 PM

which is fine to me. 

The reason I don't mind keeping the entire `userInfo[NSUnderlyingErrorKey]` is that we don't extract all the data from it, so whatever else it can offer us may be useful for debugging – IMHO it doesn't hurt to have it?

Automattic-Tracks-iOSTests/Tests/SentryExtensionTests.swift
jkmassel: Yes, though you'd have to go out of your way to do so – the default scheme is set to use the test plan, so you'd need to override the defaults to make that happen. Even so, I applied this suggestion because it never hurts to document things better 😁

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
jkmassel: Just a habit I got into when using Rust – I particularly like it for situations like this where I don't control the `SentryEvent` object – I feel like they own the `init` namespace, so I like static initializers for this case 🤷

AliSoftware: Right, got it. Kinda makes sense.

It's just that for people used to Swift, I don't think they will think about trying `SentryEvent.from ⇥` to check with autocomplete if there's a way to create a `SentryEvent` from an `NSError`. It's more idiomatic and common to type `SentryEvent  ⇥` or `SentryEvent.init ⇥` to discover all the ways to create a `SentryEvent` by relying on autocomplete, so having it as a `static func` might reduce discoverability here. But 🤷 not that big a deal.

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
AliSoftware: Yep, looks good 👍  As long as Sentry doesn't crash and instead properly use the object `description` in those cases, that's all good to me 👍

Automattic-Tracks-iOS/Crash Logging/SentryEvent+Extensions.swift
```diff
@@ -0,0 +1,33 @@
...
+        timestamp: Date = Date(),
+        extra: [String: Any] = [:]
+    ) -> Event {
+        let event = Event(level: level)
+
+        if let underlyingError = error.userInfo[NSUnderlyingErrorKey] as? NSError {
+            event.message = SentryMessage(formatted: underlyingError.debugDescription)
+        } else {
+            event.message = SentryMessage(formatted: error.debugDescription)
+        }
```
AliSoftware: Code style suggestion (take it or leave it):
```suggestion
        let baseError: NSError = (error.userInfo[NSUnderlyingErrorKey] as? NSError) ?? error
        event.message = SentryMessage(formatted: baseError.debugDescription)
```

AliSoftware (approved):
LGTM.

Still would prefer an `init` over `static func` (https://github.com/Automattic/Automattic-Tracks-iOS/pull/173#discussion_r601279101); but maybe that's just me 🤷  (and that's just a nit that shouldn't block this PR to get merged).

PR #174, opened 2021-03-18
ExPlat: revert hardcoded platform and remove system notifications

leandroalonso:
Please review using: https://github.com/wordpress-mobile/WordPress-iOS/pull/16129


momo-ozawa (approved):
👍

PR #175, opened 2021-03-26
[Tooling] Standardize podspec format

AliSoftware:
As part of paaHJt-Vl-p2, this simply standardize the podspec file so that all our pods use a similar format and structure, with similar keys and common values.

See wordpress-mobile/WordPressAuthenticator-iOS#582 for the rules I decided to follow on the format and content standardization.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,28 @@ jobs:
                        -scheme 'Automattic-Tracks-OSX' \
                        -destination 'platform=OS X,arch=x86_64' \
                        test  | tee logs/build-and-test.log | xcpretty -r junit
+  CheckVersionConsistency:
+    docker:
+      - image: cimg/base:2020.01
+    steps:
+      - checkout
+      - run:
+          name: Check pod version consistency
+          command: |
+            SRC_FILE=Automattic-Tracks-iOS/TracksConstants.m
+            PODSPEC_FILE=Automattic-Tracks-iOS.podspec
 
+            SOURCE_VERS=$(sed -n s/'^.* TracksLibraryVersion = @"\(.*\)";.*$'/'\1'/p $SRC_FILE)
+            POD_VERS=$(sed -n s/'^ *s.version *= \([^ ]*\).*$'/'\1'/p $PODSPEC_FILE | tr -d \'\")
+            
+            echo Version found in $SRC_FILE: $SOURCE_VERS
+            echo Version found in $PODSPEC_FILE: $POD_VERS
+            if [ "$SOURCE_VERS" = "$POD_VERS" ]; then
+              echo "Version values match :)"
+            else
+              echo "Version values differ! Please fix by making sure the 2 files have matching values for the version."
+              exit 1
+            fi
 workflows:
   test_and_validate:
     jobs:
@@ -67,5 +88,7 @@ workflows:
               only: /.*/
             branches:
               ignore: /.*/
+      - CheckVersionConsistency:
+          name: "Validate Version Values Consistency"
       - TestMacOS:
           name: "Test macOS"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,6 +48,7 @@ jobs:
               echo "Version values differ! Please fix by making sure the 2 files have matching values for the version."
               exit 1
             fi
+
 workflows:
   test_and_validate:
     jobs:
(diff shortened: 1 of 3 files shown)


Automattic-Tracks-iOS.podspec
```diff
@@ -1,36 +1,42 @@
...
-  spec.module_name = 'AutomatticTracks'
-
-  spec.ios.dependency 'UIDeviceIdentifier', '~> 1'
-  spec.dependency 'CocoaLumberjack', '~> 3'
-  spec.dependency 'Reachability', '~> 3'
-  spec.dependency 'Sentry', '~> 6'
-  spec.dependency 'Sodium', '>= 0.9.1'
+Pod::Spec.new do |s|
+  s.name          = 'Automattic-Tracks-iOS'
+  s.version       = File.read("Automattic-Tracks-iOS/TracksConstants.m").split("const TracksLibraryVersion = @\"").last.split("\"").first
```
jkmassel: I don't love the fact that this is as non-standard as it is (it's the only pod that works this way). WDYT about aligning it with the other projects?

jkmassel (changes requested):
Looking for feedback on this one – IMHO it'd be nice for it to be the same as all our other pods, but I understand that might not be possible

Automattic-Tracks-iOS.podspec
AliSoftware: I actually pondered about that and would love to migrate to hardcoded value here indeed.

Thing is, if we do this it could be nice to also add in the CI a step that checks that the value here matches the one in the .m to ensure they are still always in sync and we do to forget to update both at once. Will try to tackle this next week.

AliSoftware:
* Using hardcoded version in podspec in 2ea0fa5
* Added a CI check to ensure consistency in version values in 76ae9bf
  * Created an intentional inconsistency in 6659074 ==> [Confirms that the CI check fails here](https://app.circleci.com/pipelines/github/Automattic/Automattic-Tracks-iOS/538/workflows/0837345e-a90b-497e-8095-86f02c287bd9/jobs/1413)
  * Reverted the intentional mismatch in 129b4e7 ==> [Confirms that the CI passes here](https://app.circleci.com/pipelines/github/Automattic/Automattic-Tracks-iOS/537/workflows/9cf9a07f-964e-4ef7-a301-69a7c728cb0a/jobs/1418)

🎉

Automattic-Tracks-iOS.podspec
AliSoftware: @jkmassel Fixed in 2ea0fa57de1dd6c022fa60c81390802dd7d13675 and CI check added in 76ae9bf.
See also [this comment below](https://github.com/Automattic/Automattic-Tracks-iOS/pull/175?xs#issuecomment-809268052) for details.

.circleci/config.yml
```diff
@@ -26,7 +26,28 @@ jobs:
                        -scheme 'Automattic-Tracks-OSX' \
                        -destination 'platform=OS X,arch=x86_64' \
                        test  | tee logs/build-and-test.log | xcpretty -r junit
+  CheckVersionConsistency:
+    docker:
+      - image: cimg/base:2020.01
```
AliSoftware, line 31: Using the most basic of [CircleCI's NextGen conveninence images](https://circleci.com/docs/2.0/circleci-images/#next-generation-convenience-images) with just the bare minimum here because we only need to run a very simple shell script and this one is likely one of the fastest to allocate and boot.

.circleci/config.yml
```diff
@@ -26,7 +26,28 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Check pod version consistency
+          command: |
+            SRC_FILE=Automattic-Tracks-iOS/TracksConstants.m
+            PODSPEC_FILE=Automattic-Tracks-iOS.podspec
 
+            SOURCE_VERS=$(sed -n s/'^.* TracksLibraryVersion = @"\(.*\)";.*$'/'\1'/p $SRC_FILE)
+            POD_VERS=$(sed -n s/'^ *s.version *= \([^ ]*\).*$'/'\1'/p $PODSPEC_FILE | tr -d \'\")
```
AliSoftware, line 41: Note: It's easiest to have `sed` match the version _with_ the quotes (be it simple or double quotes) then have `tr` remove the quotes, than trying to use `["']` in the regex to match any kind of quotes.

This is because escaping `["']` within the single-quote that wraps the whole regex would make it both tricky to read and creates quite a mess in escape sequences required.

jkmassel (approved):
Nice! Thanks for taking care of this!

PR #176, opened 2021-03-30
ExPlat: fix refreshIfNeeded

leandroalonso:
This PR fixes a silly issue (yes, I know, shame on me) in `refreshIfNeeded`:

* It should refresh *if* the TTL has expired (< 0), not the other way around

I added a unit test to avoid regression and to ensure that the issue was properly fixed.


emilylaguna (approved):
LGTM 🚀

PR #182, opened 2021-06-16
Add device_info_app_build_configuration property to all events

koke:
Adds a new `device_info_app_build_configuration` property to all events that can be:
- Debug: for local developer builds
- Alpha: for `Release-alpha` builds, I believe we use this for Installable Builds on PRs. This requires the app integrating the library to define an `ALPHA` preprocessor macro.
- Beta: for `Release` builds uploaded to TestFlight
- Production: for `Release` builds uploaded to the App Store

Alpha releases can be configured with the following in the app's `Podfile`:
```ruby
  # Flag Alpha builds for Tracks
  # ============================
  installer.pods_project.targets.each do |target|
    next unless target.name == "Automattic-Tracks-iOS"
    target.build_configurations.each do |config|
      next unless config.name == "Release-Alpha"
      config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)', 'ALPHA=1']
    end
  end
```

I've only implemented this for Woo in https://github.com/woocommerce/woocommerce-ios/pull/4468/, so I'm not sure if integrating this in WordPress will require any changes for the WP Internal builds.

Part of https://github.com/woocommerce/woocommerce-ios/issues/4432

## To test

1. Check out https://github.com/woocommerce/woocommerce-ios/pull/4468 and run the Woo app.
2. You can set a breakpoint in `- [TracksServiceRemote sendBatchOfEvents:withSharedProperties:completionHandler:]` and ensure that the `properties` argument contains `Debug` for `device_info_app_build_configuration`.
3. Check the Tracks Live View after a few minutes, and verify that the events coming from the app include the new property.

I haven't been able to test anything other than Debug builds yet 😞


koke:
@Automattic/platform-9 I could use some help testing this in configurations other than Debug. I'm not even sure if we can test the TestFlight one without actually pushing a new update, so maybe something to watch out for in the next release cycle

mokagio:
Maybe we should setup distribution for the Tracks-Demo app that's part of the project via App Center?

ctarda (approved):
I followed the testing instructions and I can see `device_info_app_build_configuration` in properties. 

I tested only with the WCiOS app though. I think it would not hurt to test with WPiOS as well, sometime net week perhaps?

:shipit:

koke:
>  I think it would not hurt to test with WPiOS as well, sometime net week perhaps?

It is not integrated on WPiOS yet, but I'll give it a try next week

koke:
For reference, tested this on a TestFlight release and it's showing the correct value 😌 

Screen Shot 2021-06-30 at 11 08 57

astralbodies:
Thanks again for doing this, @koke!

PR #183, opened 2021-06-16
Crash logging: Add hybrid SDK helpers to allow better stack traces for other platforms

fluiddot:
This PR adds helpers that will allow us to improve the stack traces that we get when an exception is caused by hybrid SDKs, like React native for the Gutenberg editor.

**Related PRs:**
`WordPress-iOS`PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16700
`gutenberg-mobile` PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3629
`gutenberg` PR: https://github.com/WordPress/gutenberg/pull/32768

### DIFF
--- a/Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
+++ b/Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
@@ -286,6 +286,100 @@ extension CrashLogging {
     }
 }
 
+// MARK: - Helpers for hybrid SDKs
+extension CrashLogging {
+    /// Returns the options required to initialize Sentry in other platforms.
+    public func getOptionsDict() -> [String: Any] {
+        return [
+            "dsn": self.dataProvider.sentryDSN,
+            "environment": self.dataProvider.buildType,
+            "releaseName": self.dataProvider.releaseName
+        ]
+    }
+    
+    /// Return the current Sentry user.
+    /// This helper allows events triggered by other platforms to include the current user.
+    public func getSentryUserDict() -> [String: Any]? {
+        return dataProvider.currentUser?.sentryUser.serialize()
+    }
+    
+    /// Attachs the current scope to an event and returns it.
+    /// This helper allows events triggered by other platforms to include the same scope as if they would be triggered in this platform.
+    ///
+    /// - Parameters:
+    ///   - eventDict: The event object
+    public func attachScopeToEvent(_ eventDict: [String: Any]) -> [String: Any] {
+        var scope = SentrySDK.currentHub().getScope().serialize()
+        
+        // Setup tags
+        var tags = scope["tags"] as? [String: String] ?? [String: String]()
+        tags["locale"] = NSLocale.current.languageCode
+        
+        /// Always provide a value in order to determine how often we're unable to retrieve application state
+        tags["app.state"] = ApplicationFacade().applicationState ?? "unknown"
+        
+        tags["release"] = self.dataProvider.releaseName
+        
+        // Assign scope to event
+        var eventWithScope = eventDict
+        eventWithScope["tags"] = tags
+        eventWithScope["breadcrumbs"] = scope["breadcrumbs"]
+        eventWithScope["contexts"] = scope["context"]
+        
+        return eventWithScope
+    }
+    
+    /// Writes the envelope to the Crash Logging system, the envelope contains all the data required for data ingestion in Sentry.
+    /// This function is based on the original Sentry implementation for React native: https://github.com/getsentry/sentry-react-native/blob/aa4eb11415cbb73bbd0033e4f0926b539d22315b/ios/RNSentry.m#L118-L158
+    ///
+    /// - Parameters:
+    ///   - envelopeDict: The envelope object.
+    public func logEnvelope(_ envelopeDict: [String: Any]) {
+        if(JSONSerialization.isValidJSONObject(envelopeDict)) {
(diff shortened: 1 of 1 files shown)


Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -50,19 +50,27 @@ public class CrashLogging {
         return self
     }
 
+    public func shouldSendEvent() -> Bool {
+        #if DEBUG
+        return UserDefaults.standard.bool(forKey: "force-crash-logging")
+        #else
+        return !dataProvider.userHasOptedOut
+        #endif
+    }
```
fluiddot, line 59: Initially, I thought that `beforeSend` would be called in all cases but looks like that for capturing envelopes this is skipped. For this reason, I moved out this logic from `beforeSend` and expose it, this way Gutenberg can access it and determine if an event should be sent.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -278,6 +286,104 @@ extension CrashLogging {
...
+// MARK: - Helpers for hybrid SDKs
+extension CrashLogging {
+    /// Returns the options required to initialize Sentry in other platforms.
+    public func getOptionsDict() -> [String: Any] {
+        return [
+            "dsn": self.dataProvider.sentryDSN,
+            "environment": self.dataProvider.buildType,
+            "releaseName": self.dataProvider.releaseName
+        ]
+    }
```
fluiddot, line 298: In hybrid SDKs like [React Native](https://docs.sentry.io/platforms/react-native/) (used in Gutenberg), we need to initialize the SDK so I exposed the required options to match the same values we use for the native SDK.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -278,6 +286,104 @@ extension CrashLogging {
...
+
+    /// Return the current Sentry user.
+    /// This helper allows events triggered by other platforms to include the current user.
+    public func getSentryUserDict() -> [String: Any]? {
+        return dataProvider.currentUser?.sentryUser.serialize()
+    }
```
fluiddot, line 304: Similar to `attachScopeToEvent`, I exposed the Sentry user for including it in the events generated in the hybrid SDKs. Initially, I thought that the user would be part of the current scope but it's not as it's assigned to the event in the `beforeSend` callback.

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -278,6 +286,104 @@ extension CrashLogging {
...
+        tags["release"] = self.dataProvider.releaseName
+
+        // Assign scope to event
+        var eventWithScope = eventDict
+        eventWithScope["tags"] = tags
+        eventWithScope["breadcrumbs"] = scope["breadcrumbs"]
+        eventWithScope["contexts"] = scope["context"]
+
+        return eventWithScope
+    }
```
fluiddot, line 330: The scope used in the events generated in hybrid SDKs is different from the one of the native side so I expose this function to incorporate the same data we see in native crashes.

This logic serializes the current scope and includes the same extra tags added in the `beforeSend` callback, finally, a new object is created with the following data set:
- Tags
- Breadcrumbs
- Contexts

Automattic-Tracks-iOS/Crash Logging/CrashLogging.swift
```diff
@@ -278,6 +286,104 @@ extension CrashLogging {
...
+                SentrySDK.currentHub().getClient()?.store(envelope)
+            } else {
+                SentrySDK.currentHub().getClient()?.capture(envelope: envelope)
+            }
+            #endif
+        } else {
+            DDLogError("⛔️ Unable to send envelope to Sentry – envelope is not a valid JSON object.")
+        }
+    }
+}
```
fluiddot, line 385: Hybrid SDKs send the events as [envelopes](https://develop.sentry.dev/sdk/envelopes/), it's basically an object that contains all the data required to ingest an event into Sentry.

This implementation is based on the [original version of the Sentry React native SDK (version `2.4.2`)](https://github.com/getsentry/sentry-react-native/blob/a1b314781fb673becfb572a97a297b9bb8e7a849/ios/RNSentry.m#L130-L172). I'd like to note that the code reference is from version `2.4.2` because is the last version that [uses Sentry iOS SDK version `6.x`](https://github.com/getsentry/sentry-react-native/blob/a1b314781fb673becfb572a97a297b9bb8e7a849/RNSentry.podspec#L20) that should match the version we use for crash logging ([currently `6.x` from the podspec](https://github.com/Automattic/Automattic-Tracks-iOS/blob/36db74cceaf98e107e0c713eaae81eb4139c2059/Automattic-Tracks-iOS.podspec#L40)).

fluiddot:
👋 @jkmassel I'd appreciate if you could take a look at this implementation that enables the Sentry integration with React native, let me know if you're available, thank you very much for your help 🙇 !

**Related issue:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/1187.

antonis (approved):
Tested via https://github.com/WordPress/gutenberg/pull/32768#pullrequestreview-692939310 
The code changes LGTM too 🎉

fluiddot:
Surpassed by https://github.com/Automattic/Automattic-Tracks-iOS/pull/278.

PR #186, opened 2021-07-27
Add support for Swift Package Manager

bjhomer:
This PR adds support for Swift Package Manager!

Don't be scared off by the huge number of changes. The majority of the deletions are removing the old `.xcodeproj` package that is no longer needed, and the majority of the additions are in the new `Package.swift` file that describes the package structure.

A few highlights to help guide you through this PR:

### Source compatibility
This should be fully source-compatible with anyone using Tracks via Cocoapods. (Please check me on that, though!) And switching an existing project over to SPM from Cocoapods should be as simple as adding the SPM dependency and removing the Cocoapod. No code changes should be required.

### Organizational changes

When built under SPM, the single "AutomatticTracks" library has now been split into multiple separate modules that implement the various pieces of functionality
- `AutomatticTracksEvents` for reporting events
- `AutomatticExperiments` for running A/B tests using ExPlat
- `AutomatticRemoteLogging` for uploading app logs and crash logs
- `AutomatticCrashLoggingUI` for displaying crash logs in a SwiftUI view
- `AutomatticTracks` imports **all of the above**

There are also a couple other modules that just implement shared code: `AutomatticTracksModel` and `AutomatticTracksModelObjC`. These aren't very useful on their own, but it seemed useful to pull out the common code.

If you're using the library via Cocoapods, we still just expose `import AutomatticTracks` as before, so there should be no source changes needed there.

### Directory structure

The directory structure has also changed. SPM really wants all files of a single target to belong to a single directory. Because SPM can't have Swift and ObjC code in the same target, I had to make some changes anyway, and since I was doing so, I decided to bring them more in line with what SPM would usually produce. Here's a rough overview of the changes:
- Before: 
    Screen Shot 2021-07-27 at 9 16 50 AM
- After: 
    Screen Shot 2021-07-27 at 9 17 12 AM

Most of those loose files have just been deleted, as they were unnecessary.

### Other changes of note.

- The README has been updated to explain how to install using SPM.
- The "Reachability" dependency has been removed. It was using deprecated technology, which is now implemented using nw_path_monitor.
- I've commented inline on other changes, where relevant.


Sources/Event Logging/TracksLogging.m
```diff
@@ -2,9 +2,9 @@
 #import "TracksLoggingPrivate.h"
 
 DDLogLevel TracksGetLoggingLevel(void) {
-    return ddLogLevel;
+    return tracks_ddLogLevel;
```
bjhomer: `ddLogLevel` is the default that will be used by all Cocoapods installations. If a user of this library already has defined a symbol called `ddLogLevel`, then we'll get duplicate symbols. Since this one is only intended for use within tracks, I'm prefixing it here.

(I ran into this when trying to integrate this into Day One.)

Sources/Event Logging/TracksService.m
```diff
@@ -1,24 +1,32 @@
...
 @interface TracksService ()
 
 @property (nonatomic, strong) NSTimer *timer;
 @property (nonatomic, assign) BOOL timerEnabled;
 @property (nonatomic, assign) BOOL isHostReachable;
 @property (nonatomic, strong) TracksContextManager *contextManager;
 @property (nonatomic, strong) TracksDeviceInformation *deviceInformation;
-@property (nonatomic, strong) Reachability *reachability;
+@property (nonatomic, strong) nw_path_monitor_t networkMonitor;
+@property (nonatomic, strong) nw_path_t networkPath;
```
bjhomer, line 27: As I mentioned above, Reachability is deprecated, and `nw_path_monitor` is the modern way to accomplish the same functionality.

Sources/Model/ObjC/TracksDeviceInformation.m
```diff
@@ -2,7 +2,7 @@
 
 #if TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
-#import <UIDeviceIdentifier/UIDeviceHardware.h>
+@import UIDeviceIdentifier;
```
bjhomer, line 5: UIDeviceIdentifier accepted my change, and now directly supports Swift Package Manager. This is the right way to import it when using SPM.

Sources/Model/Swift/TracksUser.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+import Sentry
+
+public struct TracksUser {
```
bjhomer, line 4: This is really just a moved file, but I pulled a bunch of `extension TracksUser {}` stuff out into a separate file to avoid layering problems, so now git isn't detecting the rename. Nothing changed in the body here.

Sources/Remote Logging/Crash Logging/ApplicationFacade.swift
```diff
@@ -37,6 +44,31 @@ final class ApplicationFacade {
 
 #if os(iOS)
 
+private extension UIApplication {
+    // When compiling with Swift Package Manager, it wants us to
+    // use only extension-safe API. We still want to use UIApplication
+    // when it's available, while not using extension-unsafe API
+    // when not available. So we're going to be sneaky about getting
+    // `UIApplication.shared`, but only when it should already be safe.
```
bjhomer: At least in Xcode 13 b3, SPM packages are compiled in extension-safe mode, which prohibits calling `UIApplication.shared`. This change may be reverted in later Xcode betas, as there has been a lot of negative feedback on the topic, but this code should be safe either way.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -56,7 +62,7 @@ public class CrashLogging {
 
         #if DEBUG
         DDLogDebug("📜 This is a debug build")
-        let shouldSendEvent = UserDefaults.standard.bool(forKey: "force-crash-logging")
+        let shouldSendEvent = UserDefaults.standard.bool(forKey: "force-crash-logging") && !dataProvider.userHasOptedOut
```
bjhomer: Tests were failing without this change; the tests were trying to verify whether `userHasOptedOut` was working, but in this DEBUG branch, it wasn't ever being looked at.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -278,17 +284,30 @@ extension CrashLogging {
     }
 }
 
-// MARK: - Event Logging
-extension Event {
+internal extension TracksUser {
```
bjhomer, line 290: This is the code that used to be in TracksUser.swift. It's just moved directly, with no changes.

Sources/Remote Logging/Crash Logging/SentryClient+Shim.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+//  Created by BJ Homer on 7/27/21.
+//
+
+import Foundation
+import Sentry
+
+
+@objc
+private protocol SentryClient_InternalMethods {
```
bjhomer, line 5: This file is a Swift re-implementation of `SentryClient+Shim.m`. I mostly avoided rewriting code for this project, but in this case, we would have needed an entire separate module just for this one file.

Sources/Remote Logging/Encrypted Logs/EventLogging+Sentry.swift
```diff
@@ -49,4 +54,16 @@ extension Event {
...
+    private static let logIDKey = "logID"
+
+    var logID: String? {
+        get {
+            return self.extra?[Event.logIDKey] as? String
+        }
+        set {
+            self.extra?[Event.logIDKey] = newValue
+        }
+    }
```
bjhomer, line 67: This used to be in CrashLogging.swift. I moved it closer to where it's actually used. (At one point, I was trying to split remote logging and crash logging into separate modules. They're fairly tightly coupled, though, so I abandoned that for now. But this change was part of that.)

Tests/Tests/CrashLoggingTests.swift
```diff
@@ -70,7 +77,12 @@ class CrashLoggingTests: XCTestCase {
     func testWhenRunningOniOSThenEventsAreSentWithApplicationState() throws {
         let crashLogging = try CrashLogging(dataProvider: mockDataProvider).start()
         let event = crashLogging.beforeSend(event: Event(level: .debug))
+        #if SWIFT_PACKAGE
+        // When building for SPM, there is no active application for the tests.
+        XCTAssertEqual("unavailable", event?.tags?["app.state"])
```
bjhomer, line 82: I don't know if this is just an SPM thing, or if it's a change in recent versions of Xcode.

Tests/Tests/ObjC/TracksEventServiceTests.m
```diff
@@ -49,7 +54,7 @@ - (void)testTracksEventWithNilName {
 
 - (void)testCreateTracksEventForAliasingWordPressComUserValidData
 {
-    OCMStub([self.eventPersistenceMock persistTracksEvent:[OCMArg isNotNil]]);
+    OCMExpect([self.eventPersistenceMock persistTracksEvent:[OCMArg isNotNil]]);
```
bjhomer, line 57: This test wasn't doing anything. Now it is.

README.md
```diff
@@ -7,20 +7,58 @@ Tracks for iOS is a client library used to help track events inside of an applic
 
 ## Installation
 
-You can install the Tracks component in your app via [CocoaPods](http://cocoapods.org):
+
+You can install the Tracks component in your app via Swift Package Manager:
+
+```swift
+.package(url: "https://github.com/Automattic/Automattic-Tracks-iOS", from: "0.10.0")
```
bjhomer, line 14: I'm making the assumption that when this is merged in, we'll make a "0.10.0" tag

Scripts/build.sh
```diff
@@ -1,14 +1,4 @@
...
-XCODE_PROJECT=Automattic-Tracks-iOS.xcodeproj
-XCODE_SCHEME=Automattic-Tracks-iOS
-XCODE_SDK=iphonesimulator
-
-xcodebuild build test \
--workspace "$XCODE_WORKSPACE" \
--scheme "$XCODE_SCHEME" \
--sdk "$XCODE_SDK" \
--destination "name=iPhone SE" \
--configuration Debug | bundle exec xcpretty -c
```
jkmassel: 😍

Sources/Event Logging/TracksLogging.m
jkmassel: Thanks for dealing with this – some history – this is a bit of an ugly workaround we had to make to sync up log levels because all our dependencies use CocoaLumberjack directly.

I've long considered using a protocol that the parent app can confirm to, which would eliminate the need for a dependency on CocoaLumberjack (and really, any specific logging framework) – instead, the parent app would delegate messages to its logging framework of choice.

Not sure if that's in-scope for this, but because we're changing so much already I wonder about trying to sneak it in 😅

Sources/Remote Logging/Crash Logging/CrashLogging.swift
jkmassel: Whoa nice catch!

Sources/Remote Logging/Crash Logging/SentryClient+Shim.swift
jkmassel: Nice – thanks!

Sources/Event Logging/TracksLogging.m
bjhomer: I actually thought about doing something like that, but I've been trying to keep my changes here source-compatible with other existing clients. It sounds like this change would cause Tracks logging to not appear in the logs of existing apps, so that would be a problem.

A simple logging protocol could look something like this:

```swift
@objc protocol TracksLogging {
  func logError(_: String)
  func logWarn(_: String)
  func logInfo(_: String)
  func logDebug(_: String)
  func logVerbose(_: String)
}
```

Does something like that seem reasonable?

Ecarrion:
Great work @bjhomer!

I tried an early integration on the Woo App and these are my findings.

1. Tracks events work as expected!
Screen Shot 2021-07-29 at 10 37 29 AM

2. We end up with multiple dependencies on `CocoaLumberJack`.
I tried removing the `Cocoapods` ones, but the whole project starts to scream. I spent 30 minutes looking at it, but wasn’t able to fix it right away.

```
The project outputs warnings about this, EG:
Class DDAbstractDatabaseLogger is implemented in both /private/var/containers/Bundle/Application/43E99593-901F-4F18-9371-5E41A513FF75/WooCommerce.app/Frameworks/Networking.framework/Networking (0x10a511868) and /private/var/containers/Bundle/Application/43E99593-901F-4F18-9371-5E41A513FF75/WooCommerce.app/Frameworks/CocoaLumberjack.framework/CocoaLumberjack (0x1058d2420). One of the two will be used. Which one is undefined.
```

3. I tried generating a `crash/error` on Sentry, but I’m not sure if it is not showing because something is wrong, or because I'm on DEBUG/Local configurations.

Maybe @jkmassel could help with `2.` and `3.`? 😅

bjhomer:
@Ecarrion Thanks for checking it out! One clarification: are you adding the library via Cocoapods or via SPM in your test here? I'm _anticipating_ that nothing will have changed for users who are integrating via Cocoapods, so I'm guessing you're using SPM. Is that correct?

I'm guessing that having Tracks pull in CocoaLumberjack via an SPM dependency and having the app pull it in via pods is likely going to result in the duplicate framework issues you are seeing here. @jkmassel: we could resolve this by adding the logging protocol as you suggested. I'm happy to do that; the only caveat is that it will mean that this is not a source-compatible update, and clients of the library would have to make some changes to preserve their logging. Do you feel like that's acceptable?

My guess is that having a single app that uses both Cocoapods and SPM for dependencies is likely to run into problems any time the same dependency is being added by both systems.

Ecarrion:
@bjhomer 

> so I'm guessing you're using SPM. Is that correct?

Correct!

jkmassel:
@bjhomer 

> the only caveat is that it will mean that this is not a source-compatible update, and clients of the library would have to make some changes to preserve their logging. Do you feel like that's acceptable

I do – we're not at `1.0.0` yet on this project, so there's been several source-breaking changes already – IMHO this one is critical anyway for moving forward so I'm more than happy for us to have to deal with it now 🙂

bjhomer:
Okay, I think this is ready to review again.

README.md
jkmassel: Yep!

jkmassel:
@mokagio – tagging you for review in this, because I had to replace the work done in https://github.com/Automattic/Automattic-Tracks-iOS/pull/147 – even with the test target sandboxed, the tests ask for access to the user's `Documents` directory.

After this change, the Tracks DB will always live in the `Application Support` directory – whether inside the container, or inside `~/Library/Application Support` – thus never triggering this prompt.

The downside is that we'll lose some tracks events in the transition, but it shouldn't be very many, we should only need to do it once, and it'll only be on macOS.

Package.swift
```diff
@@ -0,0 +1,133 @@
...
+            ]),
+
+        // Xcode 12 has an issue where the first build after
+        // cleaning fails if there is a dependency that vends
+        // a binary .xcframework. We are working around this
+        // on CI by first building this "_WorkaroundSPM"
+        // target that _only_ builds that one dependency.
+        // We ignore any failures when building this target.
+        // Then we go on to build the actual product, which
+        // builds correctly.
```
mokagio, line 29: Is there a radar for this? Do you happen to have tested it on Xcode 13?

README.md
```diff
@@ -7,20 +7,66 @@ Tracks for iOS is a client library used to help track events inside of an applic
...
+
+### To upload files:
+
+1. Create an instance of `EventLogging` using `init(dataSource:delegate:)`.
+2. Call `enqueueLogForUpload(log:)` to schedule log files for uploading.
+
+
+### Logging
+
+Tracks logs about some of its activity. By default, this logging will just go to the console. If you'd like to include this logging in your own logging solution, you can assign a logging delegate:
```
mokagio: Nitpick. While the compiler would tell a user what type to use, it might be handy to suggest the delegate type here. E.g.:

```suggestion
Tracks logs about some of its activity. By default, this logging will just go to the console. If you'd like to include this logging in your own logging solution, you can create and assign a logging delegate conforming to `TracksLoggingDelegate`:
```

Sources/AutomatticTracks/AllExports.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
```
mokagio: Is this import required?

Sources/AutomatticTracks/AutomatticTracks.h
```diff
@@ -0,0 +1,5 @@
+
```
mokagio: Nitpick: do we need this empty line?

```suggestion
```

Sources/Event Logging/TracksEventService.h
```diff
@@ -1,7 +1,8 @@
 #import <Foundation/Foundation.h>
 #import "TracksEvent.h"
-#import "TracksContextManager.h"
 
+
```
mokagio: Nitpick, do we need this empty line?

```suggestion
```

Sources/Event Logging/TracksService.m
```diff
@@ -250,28 +256,31 @@ - (void)setRemoteCallsEnabled:(BOOL)remoteCallsEnabled
...
 - (void)updateDeviceInformationFromReachability
 {
-    self.deviceInformation.isWiFiConnected = self.reachability.isReachableViaWiFi;
-    self.deviceInformation.isOnline = self.reachability.isReachable;
+    if (self.networkPath != nil) {
+        nw_path_status_t pathStatus = nw_path_get_status(self.networkPath);
+        self.deviceInformation.isWiFiConnected = nw_path_uses_interface_type(self.networkPath, nw_interface_type_wifi);
+        self.deviceInformation.isOnline = (pathStatus == nw_path_status_satisfied || pathStatus == nw_path_status_satisfiable);
+    }
+    else {
```
mokagio: Nitpick:

```suggestion
    } else {
```

Sources/Model/ObjC/TracksDeviceInformation.m
```diff
@@ -235,3 +236,7 @@ - (NSString *)appBuildConfiguration
 }
 
 @end
+
+
+
+
```
mokagio:
```suggestion
```

Sources/Model/ObjC/TracksLogging.h
```diff
@@ -0,0 +1,53 @@
...
+@end
+
+
+/**
+ There's some funny layering going on here. We want to
+ expose the `TracksLogging` class publicly, and that works
+ better if we define it in Swift. If we define it in ObjC,
+ then it doesn't get re-exported when we do
+ `@_exported import AutomatticTracksModelObjC` in Swift,
+ because @_exported only re-exports Swift symbols. So then
```
mokagio:
```suggestion
 because `@_exported` only re-exports Swift symbols. So then
```

Sources/Model/ObjC/TracksLogging.h
```diff
@@ -0,0 +1,53 @@
...
+ However, because the ObjC code needs to be able to *use* the
+ logging delegate, we need to somehow get access to it from
+ a layer higher.
+
+ So what we're doing is having the ObjC code load the Swift
+ class using NSClassFromString, and then verify that it
+ conforms to the expected protocol. This is effectively
+ "reaching up" one layer to access it, but it solves our
+ symbol exporting problem.
+ */
```
mokagio, line 48: Unfortunate but neat solution. Nicely done. 🤯

Sources/Model/ObjC/UIApplication+Extensions.h
```diff
@@ -0,0 +1,18 @@
...
+
+
+#if TARGET_OS_IPHONE
+@import UIKit;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface UIApplication (SharedIfAvailable)
+
+/// Returns the shared UIApplication if not running in an app extension;
```
mokagio:
```suggestion
/// Returns the shared `UIApplication` if not running in an app extension;
```

Sources/Remote Logging/Crash Logging/ApplicationFacade.swift
```diff
@@ -9,12 +9,13 @@ import UIKit
...
 /// Though in reality, this is only currently useful for iOS. XD
 ///
+
 final class ApplicationFacade {
     /// The current state of the application (e.g. background, active, or inactive). This is
     /// generally used only for `Event` tags.
     ///
     /// - Returns: The string-representation of the state or `nil` if the current platform is
-    ///            not supported.
+    /// not supported.
```
mokagio, line 22: Out of scope TIL.

This caught my attention as I usually apply this formatting to lists in Markdown when needing to go to a new line:

```swift
/// - Returns: ....
///   not supported.
```

I was curious to see if there was any difference in Xcode (there isn't) and noticed that Xcode renders the "returns" in the docs only when the annotated element is a function 🤔 

**`- Returns` on a function**
Screen Shot 2021-09-13 at 1 37 27 pm

**`- Returns` on a computed `var`**
Screen Shot 2021-09-13 at 1 37 48 pm

---

🤷

Sources/Remote Logging/Crash Logging/ApplicationFacade.swift
```diff
@@ -23,7 +24,13 @@ final class ApplicationFacade {
             return "unavailable"
         }
 
-        return UIApplication.shared.applicationState.descriptionForEventTag
+        if let app = UIApplication.sharedIfAvailable {
+            return app.applicationState.descriptionForEventTag
+        }
+        else {
+            return "unavailable"
+        }
```
mokagio: Did you consider using nil coalescing?

```swift
return UIApplication.shardIfAvailable?.applicationState.descriptionForEventTag ?? "unavailable"
```

Sources/Remote Logging/Crash Logging/ApplicationFacade.swift
```diff
@@ -37,6 +44,31 @@ final class ApplicationFacade {
...
+        let sharedAppSelector = Selector(("sharedApplication"))
+        if let appClass = NSClassFromString("UIApplication"),
+           let performableClass = (appClass as Any) as? NSObjectProtocol,
+           performableClass.responds(to: sharedAppSelector),
+           let performResult = performableClass.perform(sharedAppSelector),
+           let app = performResult.takeUnretainedValue() as? UIApplication {
+
+            return app
+        }
+        return nil
```
mokagio: This "early" return seems like a good use case for guard let like done above. What do you think?

Sources/Remote Logging/Crash Logging/CrashLogging.swift
mokagio: 👍 

Do we need to explicitly declare that it's `internal`? Not that it's a problem, but it made me curious since `internal` should be the default access level value.

Sources/Remote Logging/Crash Logging/SentryClient+Shim.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  SentryClient+Shim.swift
+//  AutomatticTracksiOS
+//
+//  Created by BJ Homer on 7/27/21.
+//
+
```
mokagio:
```suggestion
```

I'm not sure whether it's documented anywhere, but we don't keep this info header that Xcode generates for us. Having it adds a bit of clutter to the file and all that info can be retrieved via Git anyway.

Sources/Remote Logging/Crash Logging/SentryClient+Shim.swift
```diff
@@ -0,0 +1,53 @@
...
+import Foundation
```
mokagio: I haven't tried to build this, but just as a quick checked I'd say this import is not necessary?

Sources/Remote Logging/RemoteLoggingExports.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
```
mokagio:
```suggestion
```

I _think_ we shouldn't need to explicitly import Foundation in here? Or am I missing something? 🤔

Sources/Workaround-SPM/File.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  File.swift
+//  
+//
+//  Created by BJ Homer on 7/27/21.
+//
+
```
mokagio: What do you think of replacing this with a few lines explaining why we need this workaround here?

Tests/Tests/CrashLoggingTests.swift
```diff
@@ -1,18 +1,25 @@
...
     private var mockDataProvider = MockCrashLoggingDataProvider()
 
     override func setUp() {
         mockDataProvider.sentryDSN = validDSN
+        UserDefaults.standard.setValue(true, forKey: "force-crash-logging")
     }
 
     override func tearDown() {
         mockDataProvider.reset()
+        UserDefaults.standard.removeObject(forKey: "force-crash-logging")
```
mokagio: Nitpick: we could extract the `"force-crash-logging"` string in an instance `private let` to prevent typos in the future.

Tests/Tests/CrashLoggingTests.swift
mokagio: I don't know the answer either, but I would guess it's an SPM thing, something to do with Linux compatibility maybe? When running the tests for a package with `swift test`, for example, the Simulator doesn't load. Also, in the template projects, tests are simply files in a folder, there's no mention of an application in the `Package.swift` or in the project Xcode generates to work on the package. 🤔

TracksDemo/TracksDemo Mac/TracksDemo Mac.entitlements
```diff
@@ -2,7 +2,7 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>IDEDidComputeMac32BitWarning</key>
+	<key>com.apple.security.app-sandbox</key>
```
mokagio, line 5: Because of the `.plist` structure, Git considers this a rename, but it's actually two actions:

- Add a `.entitlements` for the Mac demo app to enable sandboxing
- Remove the `IDEWorkspaceChecks.plist` file

I wonder why the checks file was removed 🤔 Is this something you did intentionally, or something Xcode automated away?

mokagio (approved):
Thank you for doing all this great work @bjhomer. Impressive.

A question for @jkmassel: this PR updates both CircleCI and Buildkite. Is it intentional? Are we waiting to remove CircleCI in a dedicated PR?

The tests now run on a sandboxed app. Ideally, I would like to test this in a non-sandboxed app before approving, but I'm running out of time today. Besides, the only client that is not sandboxed is AutoProxxy, on which we are not planning to do any change soon (which is unfortunate, but we have bigger fishes to fry 🤷 ) It might be cool to add a new build configuration for the macOS demo app, to which we could add a different entitlement file which doesn't use sandboxing. But doesn't have to happen here 👍 

---

Also, this is happening on my local copy:



🙃 

Is it just me? **I'm approving this under the assumption that it's just me.**

---

There's a few questions in which I could have checked the result myself. I wanted to do so, but between my Xcode being stuck in that resolve packages loop and GitHub refusing to scrolling somehow, I'm choosing the lazy route and simply leaving the questions for you to answer. Sorry about that 😞

Package.swift
bjhomer: Yes, it works fine on Xcode 13. When I originally put up this PR, Xcode 13 was still a few months out so I was shooting for compatibility with Xcode 12. But since the final Xcode 13 build is likely to be coming out this week, we may be able to just drop this altogether. (On the other hand, this workaround doesn't cause any problems in Xcode 13 either.)

It is mentioned as fixed in the [Xcode 13 b3 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-13-beta-release-notes):

> Fixed an issue where resolution of packages with binary dependencies failed to extract the binary archive. (77011310) (FB9085487)

TracksDemo/TracksDemo Mac/TracksDemo Mac.entitlements
bjhomer: The checks file was removed when I removed the entire `.xcworkspace` file. It's no longer needed, since `Package.swift` now serves as the project structure. You can see the [commit where it was removed here](https://github.com/Automattic/Automattic-Tracks-iOS/commit/2dd75d8806ac11ed94043c90bf256774f9e90ae3).

Sources/Workaround-SPM/File.swift
bjhomer: Depending on how soon we're willing to support Xcode 13 in CI builds, we may be able to just remove this workaround altogether. But if we need to support Xcode 12 in CI builds for a while, then yeah, a comment here would make sense.

Sources/Remote Logging/RemoteLoggingExports.swift
bjhomer: I think you're right.

Sources/Event Logging/TracksService.m
bjhomer: Is there a specific style guide I should be following here? I generally refer the style I've used here; I find it a bit easier to read. But I don't feel strongly enough to push back if there's already a consensus on doing it as you suggested.

Tests/Tests/CrashLoggingTests.swift
bjhomer: Yeah, I found that when running tests via `swift test`, a bunch of stuff didn't work because there was no simulator and no app context. I had to change the CI tests to run via `xcodebuild` still (which I didn't realize was possible at first.) So this change does fix one failure in when running tests via `swift test`, but isn't sufficient to make them fully succeed in that mode.

I'm fine taking this out if we don't feel like it's useful.

Sources/Remote Logging/Crash Logging/SentryClient+Shim.swift
bjhomer: We _are_ using things like `object.responds(to selector:)` from Foundation. That's going to be imported transitively, though, so you're right that we probably don't actually need this one.

Sources/Remote Logging/Crash Logging/ApplicationFacade.swift
bjhomer: I'll do that.

Sources/Model/ObjC/TracksLogging.h
bjhomer: I did have an Apple employee mention in a private conversation that `@_exported` is _supposed_ to export ObjC symbols as well, but that doesn't seem to be working.

Sources/AutomatticTracks/AllExports.swift
bjhomer: Nope, not required. I'll remove it.

Sources/Remote Logging/Crash Logging/ApplicationFacade.swift
bjhomer: I ended up having this implemented twice (once in Swift, once in ObjC), so I'm just going to remove this one and reuse the one that already exists at a lower layer.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
bjhomer: Yeah, `internal` is the default. I sometimes do this just to make it clear to the reader that these methods are not being exposed publicly.

Tests/Tests/CrashLoggingTests.swift
bjhomer: Yeah, I'll do that.

bjhomer:
@mokagio 
> Is it just me? I'm approving this under the assumption that it's just me.

I am not seeing that repeated package resolution thing, either in Xcode 12.5 or in Xcode 13 betas. So either it's just you, or maybe you're using a different version of Xcode than the ones I've tested on.

Package.swift
mokagio: Excellent! Thank you for the update 🙇 

In my experience in Automattic Apps Division so far, we haven't been jumping onto the latest Xcode version as soon as released because of the various projects and CI setups. So, I'd lean towards shipping with this and drop it later on.

Sources/Event Logging/TracksService.m
mokagio: Very good question. It was kind of you to adopt my suggestion. I hope it didn't come across pushy (I did prefix it with _nitpick_).

We don't have an explicit styleguide that I know of. 🤔 My suggestion comes from what I've seen around our codebases and, admittedly, my personal taste 😅 

In the long run, we'd want to have a global styleguide that SwiftLint (or similar tooling) can enforce. This is obviously low priority.

Sources/Remote Logging/Crash Logging/ApplicationFacade.swift
mokagio: Oh, right! I did notice that and wanted to check in the code, but that weird infinite SPM refresh loop side tracked me 🙃

Sources/Remote Logging/Crash Logging/SentryClient+Shim.swift
mokagio: I didn't notice that call. That's what I get for leaving comments after having only "quick checked" the code 😅

mokagio:
Thank you for taking the time to address my suggestions @bjhomer and for answering my questions 🙇 .

> So either it's just you, or maybe you're using a different version of Xcode than the ones I've tested on.

I get that behavior both on 12.5.1 and 13 beta 4 😞 It's not the first time SPM-related stuff behaves weird on my Mac, I'll file it in that same bucked for now.

TracksDemo/TracksDemo Mac/TracksDemo Mac.entitlements
mokagio: > `Package.swift` now serves as the project structure

I assume to open the project in this configuration you either open `Package.swift` or the entire project folder with Xcode. Right?

In such a case, how do you run the demo app from the Package version of the project in Xcode? 🤔 I tried adding a new scheme for it, but couldn't find a way. 

image

The only way I found was to open the `xcodeproj` in the `TracksDemo` folder. When I do that, Xcode regenerates the IDE checks file.

mokagio:
Something I noticed while doing some experiments around the `IDEWorkspaceChecks.plist` file (see [this comment](https://github.com/Automattic/Automattic-Tracks-iOS/pull/186#discussion_r707857316)) is that, with both Xcode 12.5.1 and 13 beta 4, when I open the project the `.swiftpm` folder is generated.

I'm pretty sure that's the expected behavior, or, at least, is what I've seen in other projects that depend on `Package.swift`.

The UX around that folder is a bit unclear, but we should at least ignored `.swiftpm/xcode`. Some references:

- https://forums.swift.org/t/files-created-when-opening-spm-package-in-xcode-11/28589/2
- https://github.com/apple/swift-package-manager/pull/2200
- https://github.com/apple/swift-package-manager/pull/2209
- The [default `.gitignore` configuration pertaining to `.swiftpm`](https://github.com/apple/swift-package-manager/blob/272b72dbfb65b69857acd8e046200e65fcb84b52/Sources/Workspace/InitPackage.swift#L263-L264) when creating a new package at the time of writing

The above as well is all based on how I open the code within Xcode: by opening the entire folder with [`open -a Xcode .`](https://github.com/mokagio/dotfiles/blob/72e074e95318386aa7694eaf488af0710076c654/aliases). Maybe I'm just not opening it the right way? 🤔

TracksDemo/TracksDemo Mac/TracksDemo Mac.entitlements
bjhomer: Ah, yep. Opening the .xcodeproj file directly is the expected way to run the demo project; the `Package.swift` file doesn't reference it at all. (Which makes sense, because users of the library don't need to know about it.)

If you want to just open the library code, you can double-click on the `Package.swift` file, and it will open the whole Swift Package structure in Xcode.

bjhomer: I don't object to checking in the IDEChecks file for the demo project, though.

bjhomer:
I've added `.swiftpm/xcode` to the gitignore, and I've committed the IDEWorkspaceChecks plist file.

I'm on Xcode 13 b5 instead of b4, so it's _possible_ the SPM looping behavior is related to that… but I wouldn't put my hopes on that.

Sources/Workaround-SPM/File.swift
jkmassel: We're not likely to ship Xcode 13 CI support until about the time macOS Monterey ships (typically most of the bugs are worked out by then). That may change if there are critical features we need from XC13.

jkmassel (approved):
Given the fact that Gio is good with this, I'm onboard to merge it at any time now 🙂

bjhomer:
Great, I'm gonna merge it!

PR #187, opened 2021-08-10
Add Buildkite

jkmassel:
Adds Buildkite alongside CircleCI in this project.

To test:
- Note that Buildkite CI jobs succeed
- Note that [pod publishing is successful](https://buildkite.com/wordpress-mobile/tracks-ios/builds/15)


.rubocop.yml
```diff
@@ -0,0 +1,16 @@
...
+  Exclude:
+    - 'Automattic-Tracks-iOS.podspec'
+
+# Override the maximum block length (some pods have long definitions)
+Metrics/BlockLength:
+  Max: 256
+
+# Override the maximum line length
+Layout/LineLength:
+  Max: 160
```
AliSoftware: 💄 nit: CRLF 😉 
```suggestion
  Max: 160

```

.buildkite/publish-pod.sh
```diff
@@ -0,0 +1,13 @@
...
+SLACK_WEBHOOK=$PODS_SLACK_WEBHOOK
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :cocoapods: Publishing Pod to CocoaPods CDN"
+publish_pod $PODSPEC_PATH
+
+echo "--- :slack: Notifying Slack"
+slack_notify_pod_published $PODSPEC_PATH $SLACK_WEBHOOK
```
loremattei, line 13: Out of curiosity as I'm catching up with our Buildkite setup: this seems to be the same in every project, so couldn't it be placed in a shared place?
Or you are deliberately creating this file on every repository to add some flexibility for the future?

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,73 @@
...
+    env: *common_env
+    plugins: *common_plugins
+    artifact_paths: ".build/logs/*.log"
+
+  ######################
+  # Build and Test macOS
+  ######################
+  - label: "🧪 Build and Test macOS"
+    key: "test_macos"
+    command: "build_and_test_pod 'mac test'"
```
AliSoftware, line 26: Catching up on Buildkite and I'm wondering where do those `build_and_test_pod`, `lint_pod`, etc… commands come from, aka where are they defined? (They seem different in nature from the `.sh` scripts that are in `.buildkite/` in the repo like `.buildkite/publish-pod.sh`…)

Are those built-in commands that BuildKite provide to us (couldn't find them in the buildkite docs tho)? Are those scripts of our own that are installed in the `$PATH` of all our BuildKite mac agents (couldn't find them in the buildkite-ci repo tho)?

And given that `build_and_test_pod` seems to take as a parameter just the name of a lane in the `Fastfile`, I'm curious of the benefit of having `build_and_test_pod 'mac test'` as opposed to `bundle exec fastlane mac test` directly for clarity?

AliSoftware (approved):
Config makes sense, BuildKite jobs are green, and pod publication works, so approving.

But just left some comments for my curiosity to learn more about BuildKite.

AliSoftware:
💡  While searching for where those `build_and_test_pod` commands and friends could be defined, I stumbled across [`Scripts/build.sh`](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Scripts/build.sh) which I think is a leftover from old configuration before we used Fastlane… and that we could maybe take that occasion to remove to clean things up (and get rid of the whole `Scripts/` dir)?

jkmassel:
> 💡 While searching for where those `build_and_test_pod` commands and friends could be defined, I stumbled across [`Scripts/build.sh`](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Scripts/build.sh) which I think is a leftover from old configuration before we used Fastlane… and that we could maybe take that occasion to remove to clean things up (and get rid of the whole `Scripts/` dir)?

Good call – removed in d00b13473d29bcd41a89d8d4cb100451f20023f7

.buildkite/publish-pod.sh
jkmassel: It _could_ be put in the same place, but it's somewhat abstraction-resistant.

Two reasons for this:
- Not all Pods have the same steps (some upload to our private specs repo, others don't)
- There are several variables that are referenced from the secrets storage – rather than relying on magic values being set at the lower levels of the abstraction, for the most part I'd prefer to pass them as arguments to hopefully make debugging simpler. 

I think that once we're using SwiftPM for all of our projects and can drop the private specs repo, we may be able to consolidate this a bit more easily?

.buildkite/pipeline.yml
jkmassel: These functions exist as part of this plugin: https://github.com/Automattic/bash-cache-buildkite-plugin

It's what's referred to as [a library plugin](https://github.com/buildkite-plugins/library-example-buildkite-plugin) in Buildkite parlance. They're exposed on the `$PATH`, as you say. 

The [`build_and_test_pod`](https://github.com/Automattic/bash-cache-buildkite-plugin/blob/trunk/bin/build_and_test_pod) call is a high-level action that installs gems, pods (if needed), and kicks off the specified lane – hopefully seeing the implementation helps clarify the intent a bit?

AliSoftware: Ahhh, I looked at the wrong repo (I looked at our private `buildkite-ci` one instead) 🤦 

> hopefully seeing the implementation helps clarify the intent a bit?

Yep, definitively cleared things up seeing that repo and its `bin/` dir contents, thanks!

---

Tbh I didn't even think of looking in that `bash-cache` plugin because the name of that plugin didn't make me think it would provide those helper commands – I thought it was only about caching (and didn't really wondered what things it'd actually cache tbh), not about providing higher level helper commands.

I now understand that this started as a plugin focused on caching (around gems, Pods/, etc), but it seems it has since grown to provide way more custom commands, most of those sure ending up calling something that involves cache (like `build_and_test_pod` calling `install_gems` internally), but that feels like an implementation detail to me 😅  . And even some of those don't involve caching at all (e.g. `slack_notify_pod_published`)

And it seems like there are now more helper commands there that are _not_ directly related to cache management, than commands that do… So I'm wondering if it wouldn't make sense to rename that library plugin to something like `bash-commands-buildkite-plugin` or `a8c-commands-buildkite-plugin` or something like that? (Would definitively have avoided me the confusion and missing those at least 😅 )

PR #188, opened 2021-09-20
Prevent tracks from stealing our watch connectivity session

johnnydhicks:
Ran the app and the delegate methods for our Watch Data Manager are working again!


bjhomer:
Context: Tracks was unconditionally setting the WCSession's delegate, even though Day One already had one. This was causing a loss of functionality in Day One's watch app.

Sources/Model/ObjC/WatchSessionManager.m
```diff
@@ -48,6 +54,12 @@ - (instancetype)init {
     return self;
 }
 
+- (void)setHasBeenPairedIfPossibleWithSession:(nonnull WCSession *)session {
```
bjhomer: This is failing the build because `WCSession` is only available on iOS.

4sythB (approved):

Automattic-Tracks-iOS.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'Automattic-Tracks-iOS'
-  s.version       = '0.10.0'
+  s.version       = '0.10.1-beta.1'
```
diegoreymendez, line 5: When the PR is about to get merged we increase the podpsec version, and as soon as it's merged we do a new release using this version number.  This ensures the new version is published in CocoaPods for other Apps to use.

jleandroperez: We're SO COOL that we're using SPM @diegoreymendez 😎

diegoreymendez:
@johnnydhicks - Since this is all green, I'll go ahead and merge.  Thanks for letting me push this forward!

PR #189, opened 2021-10-27
Separate out the encrypted logs target

chipsnyder:
This moves the encrypted logs target into its own package. This was done to help decrease the dependencies that come along with `AutomatticTracks` 

Specifically, this will only add `Sodium` as an additional dependency to apps such as Pocket Casts who would like to pull this in but without Sentry and the CoreData stacks of the other targets.

The additional dependencies will still be downloaded but won't be linked or bundled into the new library product. The work to not download the dependencies is partially implemented in [SE-0226](https://github.com/apple/swift-evolution/blob/master/proposals/0226-package-manager-target-based-dep-resolution.md)
 
 ----
 
 This is currently being pulled in as part of https://github.com/shiftyjelly/pocketcasts-ios/pull/3574


Package.swift
```diff
@@ -91,14 +108,22 @@ let package = Package(
...
+            ],
+            path: "Sources/Model/ObjC/Common",
             publicHeadersPath: ".",
             cSettings: [.headerSearchPath("../../Event Logging/private")]),
+        .target(
+            name: "AutomatticTracksConstantsObjC",
+            dependencies: [],
+            path: "Sources/Model/ObjC/Constants",
+            publicHeadersPath: ".",
+            cSettings: []),
```
chipsnyder, line 126: I separated this out to its own target as well since we really want `AutomatticEncryptedLogs` to be as trim as possible. Specifically, I was trying to avoid bringing along the [CoreData stack](https://github.com/Automattic/Automattic-Tracks-iOS/tree/develop/Sources/Model/ObjC/Core%20Data) but I wanted to honor the reuse of the constants in TracksConstants[.h](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Sources/Model/ObjC/TracksConstants.h)/[.m](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Sources/Model/ObjC/TracksConstants.m)

Sources/Remote Logging/EventLogging+Sentry.swift
```diff
@@ -3,11 +3,14 @@ import Sentry
...
+import AutomatticEncryptedLogs
 #endif
 
-public enum EventLoggingErrorType {
-    case fatal
-    case debug
+public extension EventLoggingDelegate {
+    func logError(_ error: Error, userInfo: [String: Any]?) {
+        CrashLogging.Internals.crashLogging?.logError(error, userInfo: userInfo)
+    }
```
chipsnyder, line 13: This might be a bit of a hack-around solution but I moved the Sentry error logging to the `AutomatticRemoteLogging` target. If a client pulls in `AutomatticRemoteLogging` nothing will have changed but if you only pull in `AutomatticEncryptedLogs` you'll need to implement `logError(_ error: Error, userInfo: [String: Any]?)` as part of your `EventLoggingDelegate` implementation

README.md
```diff
@@ -11,7 +11,7 @@ Tracks for iOS is a client library used to help track events inside of an applic
 You can install the Tracks component in your app via Swift Package Manager:
 
 ```swift
-.package(url: "https://github.com/Automattic/Automattic-Tracks-iOS", from: "0.10.0")
+.package(url: "https://github.com/Automattic/Automattic-Tracks-iOS", from: "0.11.0")
```
bjhomer: You updated the `.podspec` file to 0.10.0, but here we have 0.11.0. Should these be consistent?

chipsnyder: To be honest I'm not sure 😄 

When this was [added](https://github.com/Automattic/Automattic-Tracks-iOS/commit/654f9f1ea6b34aebcf681fbbe116c9efdd2a9877#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R13) the podspec was `0.9.1` and this was set to `0.10.0`. 

@bjhomer When you added that was the Readme just set to `0.10.0` anticipating the next release? If so I'll adjust that backdown.

bjhomer: Yes, I added 0.10.0 in anticipation that the SPM support would be part of a `0.10.0` release. It looks like we haven't actually cut that release yet, so we can probably just keep this as part of `0.10.0`. (We should probably tag that release soon, though.)

chipsnyder: 👍  Thanks for confirming. That change is reverted in d1ab547c4becc1450e36e14efd7641db1f8c00b5

chipsnyder:
@bjhomer @jkmassel would either of you have time to give this one a review this week? If not I can reach out to the dev channel.

I'd like to get it merged soon so I can keep the Pocket Casts PR moving.

bjhomer:
Yep, I’ll try to take a look tomorrow.

bjhomer (approved):
The changes here look reasonable to me. The majority of the changes just deal with the package structure, and I've verified that Day One still builds with the new structure, so that part's fine.

The entanglement between encrypted logs and remote logs is something I struggled with as well when I was first setting up SPM. In the end I didn't have a use case for separating them, so I decided to just keep them together, but since it seems you have a use case now, separating them is great.

The way you've pulled apart the dependency here (using a logging delegate) seems acceptable to me.

PR #190, opened 2021-12-10
[WIP] ☠️ Panic Delegation

jleandroperez:
### Details:
[TBD]

### Testing:
[TBD]

### DIFF
--- a/Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
+++ b/Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
@@ -49,7 +49,7 @@
         if (![_persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType configuration:nil URL:storeURL options:nil error:&error]) {
             // Replace this with code to handle the error appropriately.
             // abort() causes the application to generate a crash log and terminate. You should not use this function in a shipping application, although it may be useful during development.
-            TracksLogError(@"Unresolved error %@, %@", error, [error userInfo]);
+            TracksLogPanic(@"[Tracks] PersistentStore Error: %@, %@", error, [error userInfo]);
             abort();
         }
     }
@@ -81,7 +81,7 @@
     // It seems safe not to handle this error because Application Support should always be
     // available and one should always be able to create a folder in it
     if (error != nil) {
-        TracksLogError(@"Failed to create folder for %@ in Application Support: %@, %@", bundleIdentifier, error, [error userInfo]);
+        TracksLogPanic(@"[Tracks] ApplicationSupport Creation for %@ Error: %@, %@", bundleIdentifier, error, [error userInfo]);
         abort();
     }
 
--- a/Sources/Model/Swift/TracksLogging.swift
+++ b/Sources/Model/Swift/TracksLogging.swift
@@ -51,6 +51,10 @@ public class TracksLogging: NSObject, TracksLoggingConfiguration {
 
 private class DefaultTracksLogging: NSObject, TracksLoggingDelegate {
 
+    func logPanic(_ str: String) {
+        print("Tracks <☠️>: \(str)")
+    }
+
     func logError(_ str: String) {
         print("Tracks <E>: \(str)")
     }
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -178,6 +178,10 @@ public extension CrashLogging {
             return
         }
 
+        try submitSerializedEvents(payload: requestBody, callback: callback)
+    }
+
+    private func submitSerializedEvents(payload: Data, callback: @escaping ((Result<Bool, Error>) -> Void)) throws {
         let dsn = try SentryDsn(string: dataProvider.sentryDSN)
         guard let authString = dsn.getAuthString() else {
             throw Errors.unableToConstructAuthStringError
@@ -185,7 +189,7 @@ public extension CrashLogging {
 
         var request = URLRequest(url: dsn.getEnvelopeEndpoint())
         request.httpMethod = "POST"
-        request.httpBody = requestBody
+        request.httpBody = payload
         request.addValue(authString, forHTTPHeaderField: "X-Sentry-Auth")
 
         URLSession.shared.dataTask(with: request) { (responseData, urlResponse, error) in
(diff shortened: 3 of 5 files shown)


Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
```diff
@@ -49,7 +49,7 @@ - (NSPersistentStoreCoordinator *)persistentStoreCoordinator {
         if (![_persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType configuration:nil URL:storeURL options:nil error:&error]) {
             // Replace this with code to handle the error appropriately.
             // abort() causes the application to generate a crash log and terminate. You should not use this function in a shipping application, although it may be useful during development.
-            TracksLogError(@"Unresolved error %@, %@", error, [error userInfo]);
+            TracksLogPanic(@"[Tracks] PersistentStore Error: %@, %@", error, [error userInfo]);
```
MichelleTessier, line 52: See the Day One PR for my comment about naming this differently.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -178,14 +178,18 @@ public extension CrashLogging {
             return
         }
 
+        try submitSerializedEvents(payload: requestBody, callback: callback)
```
MichelleTessier, line 181: What does this do?

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -231,6 +235,59 @@ public extension CrashLogging {
         }
     }
 
+    /// Sends an `Event` immediately and triggers a callback on completion
+    func logMessageImmediately(_ message: String, properties: [String: Any]? = nil, level: SentryLevel = .info, callback: @escaping (Result<Bool, Error>) -> Void) throws {
```
MichelleTessier, line 239: Is this being called anywhere currently?

Sources/Model/Swift/TracksLogging.swift
```diff
@@ -51,6 +51,10 @@ public class TracksLogging: NSObject, TracksLoggingConfiguration {
 
 private class DefaultTracksLogging: NSObject, TracksLoggingDelegate {
 
+    func logPanic(_ str: String) {
+        print("Tracks <☠️>: \(str)")
```
MichelleTessier, line 55: Does it make sense for all platforms for logPanic to just call `func logMessageAndWait(_ message: String, properties: [String: Any]? = nil, level: SentryLevel = .info) throws`? If so, could this just directly call `func logMessageAndWait(_ message: String, properties: [String: Any]? = nil, level: SentryLevel = .info) throws` instead of having the platform do it?

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -231,6 +235,59 @@ public extension CrashLogging {
...
+
+     - Parameters:
+     - error: The error object
+     - userInfo: A dictionary containing additional data about this error.
+     - level: The level of severity to report in Sentry (`.error` by default)
+    */
+    func logMessageAndWait(_ message: String, properties: [String: Any]? = nil, level: SentryLevel = .info) throws {
+        let semaphore = DispatchSemaphore(value: 0)
+
+        var networkError: Error?
```
MichelleTessier, line 270: What's the reasoning for this blocking the thread?

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
```diff
@@ -81,7 +81,7 @@ - (NSURL *)applicationSupportURLForAppWithBundleIdentifier:(NSString *)bundleIde
     // It seems safe not to handle this error because Application Support should always be
     // available and one should always be able to create a folder in it
     if (error != nil) {
-        TracksLogError(@"Failed to create folder for %@ in Application Support: %@, %@", bundleIdentifier, error, [error userInfo]);
+        TracksLogPanic(@"[Tracks] ApplicationSupport Creation for %@ Error: %@, %@", bundleIdentifier, error, [error userInfo]);
```
MichelleTessier, line 84: Should we remove the abort below? It feels like The app doesn't need to crash just because Tracks could not be initialized?

Sources/Model/Swift/TracksLogging.swift
jleandroperez: It would absolutely do. But the problem is... I tried, and it's not possible (or at least I've been unable!).

TracksLogging lives in a Swift Package that's not the same as the one wrapping up Sentry. Adding any kind of import for the `AutomatticRemoteLogging` package results in a compiler error (circular dependencies).

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
jleandroperez: Apologies!! couldn't find anything related to this. To which PR you refer to? (checked on [this one](https://github.com/bloom/DayOne-iOS/pull/17932) but has no logging comments?).

Thank you!!

Sources/Remote Logging/Crash Logging/CrashLogging.swift
jleandroperez: I've extracted this method from [logErrorsImmediately](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Sources/Remote%20Logging/Crash%20Logging/CrashLogging.swift#L158) (develop branch).

Essentially, it's hitting the Sentry endpoint, **right away**, via HTTP.

This is done this way, so that the crashlog is not lost (in the low disk space eventuality). Plus this allows us to submit the crashlog synchronously.

OR, in the `Log Panic` mechanism scenario, we get to report Sentry Events, synchronously.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
jleandroperez: Extracted from [logErrorAndWait](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Sources/Remote%20Logging/Crash%20Logging/CrashLogging.swift#L210).

In the `Panic Log` scenario, we want errors to be reported synchronously, in order to:

- Avoid errors storing the crashlog in the filesystem (and thus, loosing the metadata)
- Immediately hit Sentry's endpoint

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
jleandroperez: I would keep the abort() call, since ... Tracks will not be functional (and thus, in an inconsistent state).

If we intend to remove the abort invocation, we should, then, build (Something similar) to a failable initializer. Otherwise the apps using this will not be aware the internal state is broken.

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
MichelleTessier: Ah, this comment: https://github.com/bloom/DayOne-iOS/pull/17932#discussion_r766954441. 
I just meant naming it `logToSentry` or `logToErrorReportingSystem` or something like that instead of `TracksLogPanic`

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
MichelleTessier: Okay, I don't have strong feelings on it I guess? @bjhomer , do you have thoughts?

Sources/Model/Swift/TracksLogging.swift
MichelleTessier: Ahhhh, I see. Ugh, sorry, I see now why this is nasty....

Sources/Remote Logging/Crash Logging/CrashLogging.swift
MichelleTessier: Mmmmm, wait so where is `submitSerializedEvents(payload: requestBody, callback: callback)` defined?

MichelleTessier: Ah, okay, I see what's happening, thanks.

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
bjhomer: I agree that we shouldn't kill the entire app. A throwing initializer would probably be appropriate.

MichelleTessier (approved):
My final thoughts:
1) I'm approving because I think this is better than what we have, which is a crash with no information .
2) I'd recommend @bjhomer to look at this and see if he has any ideas on the circular, delegate to iOS, iOS delegates back to Tracks situation, as he has more experience with Tracks and specifically how we broke it up.
3) I'd recommend someone with more Tracks experience review this.
4) I'd still probably lean towards replacing the abort() with something that doesn't crash in production builds. But I will defer to @bjhomer on this decision.

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
MichelleTessier: Eh, I've changed my mind, and I kind of like the panic name. That is basically what is going on if you call this.

jleandroperez:
Thank you ladies and gentleman!!

Closing in favor of #191 !!

PR #191, opened 2021-12-10
☠️ Fatal Scenario: Extra Metadata

jleandroperez:
### Details:
In this PR we're swapping `abort()` with an actual NSException (with extra payload), to be thrown whenever there are problems initializing the NSPersistentStoreCoordinator (or the ApplicationSupport folder cannot be created).

Our goal is to add extra metadata, that might help us track down a a crasher affecting the D1 app.

[Sample of how this looks like in Sentry](https://sentry.io/organizations/a8c/issues/2857939405/events/0996f6bb520b4f09a42829418d2424d2/?project=5959376)

cc @jkmassel WDYT sir?

### DIFF
--- a/Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
+++ b/Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
@@ -1,6 +1,10 @@
 #import "TracksContextManager.h"
 #import "TracksLogging.h"
 
+
+NSString *const TracksApplicationSupportException   = @"TracksApplicationSupportException";
+NSString *const TracksPersistentStoreException      = @"TracksPersistentStoreException";
+
 @implementation TracksContextManager
 
 #pragma mark - Core Data stack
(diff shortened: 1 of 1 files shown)


Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
```diff
@@ -82,12 +93,25 @@ - (NSURL *)applicationSupportURLForAppWithBundleIdentifier:(NSString *)bundleIde
...
+            abort();
+        }
     }
 
     return folder;
 }
 
+- (BOOL)shouldThrowUponFatalCondition {
+    // We consider it safe to Throw (whenever a Fatal Condition arises) whenever there is no global
+    // NSException handler set
```
MichelleTessier, line 111: Whoa, I am no objc exception expert! Can you give me a rundown of how this is working @jleandroperez ? What is `NSGetUncaughtExceptionHandler()`?

Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
```diff
@@ -82,12 +93,25 @@ - (NSURL *)applicationSupportURLForAppWithBundleIdentifier:(NSString *)bundleIde
     // available and one should always be able to create a folder in it
     if (error != nil) {
         TracksLogError(@"Failed to create folder for %@ in Application Support: %@, %@", bundleIdentifier, error, [error userInfo]);
-        abort();
+
+        if (self.shouldThrowUponFatalCondition) {
+            @throw [NSException exceptionWithName:TracksApplicationSupportException
```
MichelleTessier, line 98: Do these get caught somewhere? I need a rundown on objc exceptions @bjhomer.

bjhomer (approved):
This looks good to me. I normally don't love using ObjC exceptions, but in this case we're still intending to crash; we're just doing so with a bit more context.

jleandroperez:
Thank you BJ!!

PR #195, opened 2022-01-26
Merge 0.10.0, 0.10.1, and 0.11.0 into `trunk`

mokagio:
I presume we'll soon delete `trunk` and rename `develop` into `trunk`, but if I recall correctly, we want to have everything in sync and up-to-date before doing so.

I don't know why the macOS tests are failing on CircleCI. I've decided to file it as a CircleCI issue and not spend more time on it because we'll remove CircleCI in favor of Buildkite soon™


AliSoftware (approved):
The fact that the macOS tests exit with code 65 also makes me think of a fluke. In any case it's not the job of this PR to fix it if the failure was genuine, and as you said we'll drop CircleCI soon™ so if that's indeed a CircleCI fluke it will solve itself with the Buildkite migration, so I'm ok to merge this as-is.

(I think Oguz would have re-sync'd the `trunk` branch with `develop` before doing the migration on this repo anyway, but no harm in doing that sync now too here 🙃 )

oguzkocer:
> (I think Oguz would have re-sync'd the trunk branch with develop before doing the migration on this repo anyway, but no harm in doing that sync now too here 🙃 )

For the `develop` -> `trunk` migration, instead of syncing current `develop` with current `trunk`, we verify that there isn't anything in `trunk` that's not in `develop`, delete `trunk` and then rename `develop` to `trunk` which is easier and safer. (the PR base branches, branch protection rules etc will be automatically updated this way) So, this PR didn't make a practical difference, but I think it still has historical significance as it followed our practice before the transition.

Nevertheless, I've removed `trunk` and renamed `develop` to `trunk` now, so this step shouldn't be necessary anymore.

PR #199, opened 2022-03-09
Ensure network monitor path update handler is called on start

mokagio:
Earlier today, @yaelirub reported an unusual spike in events logged with the device offline in WordPress iOS starting from 19.2. 

Together with @twstokes, the tracked it to the Tracks [update to version 0.11.0](https://github.com/wordpress-mobile/WordPress-iOS/commit/ddaa3f62f0f887b8b64497c6281f013112c2ae5e) and the new network reachability logic from #186. Tanner noticed that, at launch, the network path was `nil`, resulting in [this branch](https://github.com/Automattic/Automattic-Tracks-iOS/blob/7fc4afe72cd1539575135891d90bef26ee32106c/Sources/Event%20Logging/TracksService.m#L264-L265) of the network properties logging conditional running. 

I experienced the same behavior when running the TracksDemo app from this repo ([microapps](https://increment.com/mobile/microapps-architecture/) FTW!).

The behavior I experienced was that the block meant to receive network path updates was not called when the monitor started, even though that's the expectation set by the documentation.

I did some googling but nobody reported the issue. I stumbled onto [this post](https://msolarana.netlify.app/2018/08/04/monitoring-network-changes/) by @madsolar8582 that also explicitly sets a queue on which to dispatch the update block execution. I tried it out and it did the trick 🎉 

**To test**

On `trunk`, put a breakpoint in the update block: https://github.com/Automattic/Automattic-Tracks-iOS/blob/7fc4afe72cd1539575135891d90bef26ee32106c/Sources/Event%20Logging/TracksService.m#L323

Notice that it's not hit on start, but it is if you change you network interface, for example by turning the Wi-Fi on.

Repeat on this branch and notice that the block is called when the demo app starts (fix) and also when the network changes (no regressions).

Alternatively, use `updateDeviceInformationFromReachability` as the inspection point for the breakpoint. On `trunk` here: [updateDeviceInformationFromReachability](https://github.com/Automattic/Automattic-Tracks-iOS/blob/7fc4afe72cd1539575135891d90bef26ee32106c/Sources/Event%20Logging/TracksService.m#L323). If you do that, be mindful that it gets called as part of `TracksService` init, and will have `self.networkPath` `nil` there because the monitoring hasn't been setup yet.

### DIFF
--- a/Sources/Event Logging/TracksService.m
+++ b/Sources/Event Logging/TracksService.m
@@ -320,7 +320,7 @@ NSString *const USER_ID_ANON = @"anonId";
     self.networkMonitor = nw_path_monitor_create();
 
     __weak typeof(self) weakSelf = self;
-    nw_path_monitor_set_update_handler(self.networkMonitor, ^(nw_path_t  _Nonnull path) {
+    nw_path_monitor_set_update_handler(self.networkMonitor, ^(nw_path_t _Nonnull path) {
         [weakSelf networkPathChanged:path];
     });
     nw_path_monitor_start(self.networkMonitor);
--- a/Sources/Event Logging/TracksService.m
+++ b/Sources/Event Logging/TracksService.m
@@ -327,8 +327,8 @@ NSString *const USER_ID_ANON = @"anonId";
     // Without one, the update handler doesn't get called when the monitor starts.
     dispatch_queue_attr_t attrs = dispatch_queue_attr_make_with_qos_class(
                                                                           DISPATCH_QUEUE_SERIAL,
-                                                                          QOS_CLASS_UTILITY,
-                                                                          DISPATCH_QUEUE_PRIORITY_DEFAULT
+                                                                          0,
+                                                                          0 // The relative priority within the QOS class. Can range from 0 to -15.
                                                                           );
     self.networkMonitorQueue = dispatch_queue_create("com.automattic.tracks.network.monitor", attrs);
     nw_path_monitor_set_queue(self.networkMonitor, self.networkMonitorQueue);
--- a/Sources/Event Logging/TracksService.m
+++ b/Sources/Event Logging/TracksService.m
@@ -317,8 +317,9 @@ NSString *const USER_ID_ANON = @"anonId";
         return;
     }
 
-    __weak typeof(self) weakSelf = self;
     self.networkMonitor = nw_path_monitor_create();
+
+    __weak typeof(self) weakSelf = self;
     nw_path_monitor_set_update_handler(self.networkMonitor, ^(nw_path_t  _Nonnull path) {
         [weakSelf networkPathChanged:path];
     });
--- a/Sources/Event Logging/TracksService.m
+++ b/Sources/Event Logging/TracksService.m
@@ -319,10 +319,10 @@ NSString *const USER_ID_ANON = @"anonId";
 
     __weak typeof(self) weakSelf = self;
     self.networkMonitor = nw_path_monitor_create();
-    nw_path_monitor_set_update_handler(_networkMonitor, ^(nw_path_t  _Nonnull path) {
+    nw_path_monitor_set_update_handler(self.networkMonitor, ^(nw_path_t  _Nonnull path) {
         [weakSelf networkPathChanged:path];
     });
-    nw_path_monitor_start(_networkMonitor);
+    nw_path_monitor_start(self.networkMonitor);
 }
 
 - (void)stopNetworkMonitor {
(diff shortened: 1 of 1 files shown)


Sources/Event Logging/TracksService.m
```diff
@@ -317,12 +320,24 @@ - (void)startNetworkMonitor {
...
     self.networkMonitor = nw_path_monitor_create();
-    nw_path_monitor_set_update_handler(_networkMonitor, ^(nw_path_t  _Nonnull path) {
+
+    // Create and set a queue for where the monitor can execute it's event handlers.
+    //
+    // Without one, the update handler doesn't get called when the monitor starts.
+    dispatch_queue_attr_t attrs = dispatch_queue_attr_make_with_qos_class(
+                                                                          DISPATCH_QUEUE_SERIAL,
+                                                                          QOS_CLASS_UTILITY,
+                                                                          DISPATCH_QUEUE_PRIORITY_DEFAULT
```
bjhomer: The docs indicate that the third parameter to `dispatch_queue_attr_make_with_qos_class` is a relative priority within the QOS class, and can range from 0 to -15. Usually, we'd want to pass `0`, unless we specifically know we want to set it to a lower priority than some other value.

As it happens, `DISPATCH_QUEUE_PRIORITY_DEFAULT` happens to be equal to 0, but it's semantically the wrong thing to be passing here, because this parameter doesn't take a `dispatch_queue_priority_t` at all; it just wants an `int`.

I'd recommend replacing this with `0` explicitly. That's what I've seen in most of the Apple sample code that uses this API. ([One such example here](https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/PrioritizeWorkAtTheTaskLevel.html)).

Though… see below; we may be able to drop this call altogether.

Sources/Event Logging/TracksService.m
```diff
@@ -317,12 +320,24 @@ - (void)startNetworkMonitor {
...
+                                                                          );
+    self.networkMonitorQueue = dispatch_queue_create("com.automattic.tracks.network.monitor", attrs);
```
bjhomer: Apple's [docs](https://developer.apple.com/documentation/dispatch/dispatch_queue) (and [various Apple engineers](https://mjtsai.com/blog/2021/03/16/underused-and-overused-gcd-patterns/)) recommend using `dispatch_queue_create_with_target` instead to create a serial queue that will target one of the existing concurrent views, rather than creating a new root queue. Creating a root queue will usually result in a permanent thread dedicated to this queue, which is definitely overkill for this particular use case.

Bonus: if we use the `target` API, then I don't think we need to create the `attrs` above at all.

I really wish this were documented better, but it's not. It's all in WWDC videos and Twitter lore from Apple engineers.



```suggestion
    dispatch_queue_t utilityQueue = dispatch_get_global_queue(QOS_CLASS_UTILITY, 0);
    self.networkMonitorQueue = dispatch_queue_create_with_target("com.automattic.tracks.network.monitor", DISPATCH_QUEUE_SERIAL, utilityQueue);
```

bjhomer (changes requested):

bjhomer:
Thanks for catching this. Sorry I missed it earlier!

Sources/Event Logging/TracksService.m
mokagio: @bjhomer, I tried this approach but it behaved as before: When the demo app launches, the update handler block is not called.

I pushed that version to a dedicated branch if you want to play with it https://github.com/Automattic/Automattic-Tracks-iOS/commit/2ca8b3bf817d1aa437087f40f0c8a36512201961 .

In the meantime, I applied your [other suggestion](https://github.com/Automattic/Automattic-Tracks-iOS/pull/199#discussion_r822667556) and verified it works following the manual script from the PR description.

What do you think of merging this version in, so we can ship the fix in WordPress 19.5, which is scheduled for the coming Monday, and then figure out why the code that Apple recommends doesn't work?

Thanks!

bjhomer: @mokagio I just checked out 2ca8b3b. It looks like it's missing a call to `network_monitor_set_queue`, which is probably why it wasn't working. I added in that call and verified that it does hit the breakpoint upon launch.

If you'd rather merge in this version, though, I'm willing to accept that.

bjhomer (dismissed):
I commented about how to make my proposed version work, but I'm okay with this version if you'd prefer it.

Sources/Event Logging/TracksService.m
mokagio: @bjhomer 🤦‍♂️... I had a feeling I missed something, but couldn't put my finger on it.

I incorporated your suggestion and verified it worked, as you said 🎉 Thanks!

Given my track record in this PR, I dismissed your review and asked for a new one. Much appreciated 🙇‍♂️

bjhomer (approved):
This looks good to me!

PR #201, opened 2022-03-23
Optimized the logic that increments the upload retry count.

diegoreymendez:
This PR attempts to address performance issues we're seeing where calls to `-[TracksEventService incrementRetryCountForEvents:]` are blocking the main thread.

Here's an example of the first part of the sample file:

```
Call graph:
    2164 Thread_6960953: Main Thread   DispatchQueue_<multiple>
    + 2164 start  (in dyld) + 520  [0x1028dd0f4]
    +   2164 NSApplicationMain  (in AppKit) + 1064  [0x1b174f088]
    +     2164 -[NSApplication run]  (in AppKit) + 596  [0x1b177d9b4]
    +       2164 -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:]  (in AppKit) + 1332  [0x1b178b8a4]
    +         2164 _DPSNextEvent  (in AppKit) + 844  [0x1b178d000]
    +           2164 _BlockUntilNextEventMatchingListInModeWithFilter  (in HIToolbox) + 72  [0x1b77cea9c]
    +             2164 ReceiveNextEventCommon  (in HIToolbox) + 552  [0x1b77cecdc]
    +               2164 RunCurrentEventLoopInMode  (in HIToolbox) + 292  [0x1b77cef68]
    +                 2164 CFRunLoopRunSpecific  (in CoreFoundation) + 600  [0x1aec36734]
    +                   2164 __CFRunLoopRun  (in CoreFoundation) + 2540  [0x1aec378c0]
    +                     2164 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__  (in CoreFoundation) + 16  [0x1aec79f00]
    +                       2164 _dispatch_main_queue_callback_4CF  (in libdispatch.dylib) + 944  [0x1ae9bd0ac]
    +                         2164 _dispatch_client_callout  (in libdispatch.dylib) + 20  [0x1ae9aebac]
    +                           2164 _dispatch_call_block_and_release  (in libdispatch.dylib) + 32  [0x1ae9ace60]
    +                             2164 __33-[TracksService sendQueuedEvents]_block_invoke (in DOCore) (TracksService.m:0)
    +                               2164 -[TracksEventService incrementRetryCountForEvents:] (in DOCore) (TracksEventService.m:0)
    +                                 2164 -[TracksEventPersistenceService incrementRetryCountForEvents:] (in DOCore) (TracksEventPersistenceService.m:108)
    +                                   2164 -[NSManagedObjectContext performBlockAndWait:]  (in CoreData) + 280  [0x1b4cb0380]
    +                                     2164 _dispatch_lane_barrier_sync_invoke_and_complete  (in libdispatch.dylib) + 56  [0x1ae9bde00]
    +                                       2164 _dispatch_client_callout  (in libdispatch.dylib) + 20  [0x1ae9aebac]
    +                                         2164 developerSubmittedBlockToNSManagedObjectContextPerform  (in CoreData) + 156  [0x1b4cb047c]
    +                                           2164 __62-[TracksEventPersistenceService incrementRetryCountForEvents:]_block_invoke (in DOCore) (TracksEventPersistenceService.m:0)
    +                                             2163 -[TracksEventPersistenceService findTracksEventCoreDataWithUUID:] (in DOCore) (TracksEventPersistenceService.m:0)
    +                                             ! 1942 -[NSManagedObjectContext executeFetchRequest:error:]  (in CoreData) + 3224  [0x1b4cb303c]
    +                                             ! : 1691 -[NSComparisonPredicate evaluateWithObject:substitutionVariables:]  (in Foundation) + 216  [0x1afb12e84]
    +                                             ! : | 890 -[NSFunctionExpression expressionValueWithObject:context:]  (in Foundation) + 648  [0x1afb131f8]
    +                                             ! : | + 214 +[_NSPredicateUtilities _predicateEnforceRestrictionsOnTarget:forComponentName:]  (in Foundation) + 292  [0x1afd19abc]
    +                                             ! : | + ! 195 -[NSConstantDictionary objectForKey:]  (in CoreFoundation) + 316  [0x1aec258cc]
    +                                             ! : | + ! : 192 bsearch  (in libsystem_c.dylib) + 68  [0x1aea3944c]
    +                                             ! : | + ! : | 153 comparisonUsingOrderingForStringKeys  (in CoreFoundation) + 84  [0x1aed22364]
```

## Changes:

1. The logic that increments the retry count no longer tries to run synchronously.  This was apparently locking the main thread in DayOne under certain circumstances.  Additionally if you look at the previous code you'll notice running it synchronously wasn't really being used to serialize execution.
2. The logic that queries the events we need to increase the retry count now runs in batches of 500 events at a time.
3. I've added unit tests that should help validate the changes.
4. Because implementing this code in Swift was infinitely easier, I opened a path to incrementally migrate the `AutomatticTracksEvents` module to Swift (`AutomatticTracksEventsForSwift`).  Once all code has been migrated, simply removing the ObjC module and renaming the Swift module should do. 

### Review the unit tests:

Reviewing the unit tests and trying to find flaws in them is important because they're automatically validating our expectations.

### Testing in Day One: 

1. Import the Swift Package locally into DOCore (this should replace the remote dependency) or update the remote dependency to use this branch.
2. Place a breakpoint in `TrackEventPersistenceService.swift:25`.
3. Disconnect your device from internet. 
4. Run Day One, and make sure event logging is enabled.
5. Once the breakpoint hits, debug the method being run and make sure it increases the retry count for the specified events correctly.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -60,10 +60,24 @@ let package = Package(
         .target(
             name: "AutomatticTracksEvents",
             dependencies: ["AutomatticTracksModel",
-                           "AutomatticExperiments"],
+                           "AutomatticExperiments",
+                           "AutomatticTracksEventsForSwift"],
             path: "Sources/Event Logging",
             publicHeadersPath: ".",
             cSettings: [.headerSearchPath("../Model/ObjC")]),
+
+        // Reporting events to the Tracks service
+        //
+        // This module offers a convenience incremental migration path
+        // from ObjC to Swift for AutomatticTracksEvents.
+        // Once all of the code is migrated to Swift we can just remove
+        // AutomatticTracksEvents and rename this module to AutomatticTracksEvents
+        //
+        .target(
+            name: "AutomatticTracksEventsForSwift",
+            dependencies: ["AutomatticTracksModel"],
+            path: "Sources/Event Logging (Swift)"
+        ),
 
         // Uploading app logs and crash logs
         .target(
--- a/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
+++ b/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
@@ -27,10 +27,10 @@ public class TrackEventPersistenceService: NSObject {
             event.uuid.uuidString
         }
 
-        for startIndex in stride(from: 0, to: uuidStrings.count, by: Self.incrementRetryCountBatchSize) {
-            let isLastBatch = startIndex + Self.incrementRetryCountBatchSize >= uuidStrings.count
-
-            managedObjectContext.perform {
+        managedObjectContext.perform {
+            for startIndex in stride(from: 0, to: uuidStrings.count, by: Self.incrementRetryCountBatchSize) {
+                
+                let isLastBatch = startIndex + Self.incrementRetryCountBatchSize >= uuidStrings.count
                 let results: [TracksEventCoreData]
                 let count = min(uuidStrings.count - startIndex, Self.incrementRetryCountBatchSize)
                 let uuidStringsBatch = Array(uuidStrings[startIndex ..< startIndex + count])
--- a/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
+++ b/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
@@ -10,6 +10,7 @@ import CoreData
 @objc
 public class TrackEventPersistenceService: NSObject {
     
+    private static let incrementRetryCountBatchSize = 500
     private let managedObjectContext: NSManagedObjectContext
     
     @objc
@@ -26,14 +27,12 @@ public class TrackEventPersistenceService: NSObject {
             event.uuid.uuidString
         }
         
-        let batchSize = 500
-        
-        for startIndex in stride(from: 0, to: uuidStrings.count, by: batchSize) {
-            let isLastBatch = startIndex + batchSize >= uuidStrings.count
+        for startIndex in stride(from: 0, to: uuidStrings.count, by: Self.incrementRetryCountBatchSize) {
+            let isLastBatch = startIndex + Self.incrementRetryCountBatchSize >= uuidStrings.count
             
             managedObjectContext.perform {
                 let results: [TracksEventCoreData]
-                let count = min(uuidStrings.count - startIndex, batchSize)
+                let count = min(uuidStrings.count - startIndex, Self.incrementRetryCountBatchSize)
                 let uuidStringsBatch = Array(uuidStrings[startIndex ..< startIndex + count])
                 
                 do {
--- a/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
+++ b/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
@@ -51,6 +51,10 @@ public class TrackEventPersistenceService: NSObject {
                     return
                 }
                 
+                if results.count != count {
+                    TracksLogError("Not all provided events were found in the persistence layer. This signals a possible logical error in the tracking, persistence and retry-count-incrementing code.  Please review.")
+                }
+                
                 for event in results {
                     event.retryCount = event.retryCount.intValue + 1 as NSNumber
                 }
--- a/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
+++ b/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
@@ -22,7 +22,7 @@ public class TrackEventPersistenceService: NSObject {
 
     /// Increments the retry count for the specified events in batches of
     @objc
-    public func incrementRetryCountForEvents(_ tracksEvents: [TracksEvent], onComplete completion: ((Error?) -> Void)?) {
+    public func incrementRetryCountForEvents(_ tracksEvents: [TracksEvent], onComplete completion: (() -> Void)?) {
         let uuidStrings = tracksEvents.map { event in
             event.uuid.uuidString
         }
@@ -41,7 +41,7 @@ public class TrackEventPersistenceService: NSObject {
                     TracksLogError("Error while finding track events: \(String(describing: error))")
 
                     if isLastBatch {
-                        completion?(error)
+                        completion?()
                     }
 
                     return
@@ -58,7 +58,7 @@ public class TrackEventPersistenceService: NSObject {
                 self.saveManagedObjectContext()
 
                 if isLastBatch {
-                    completion?(nil)
+                    completion?()
                 }
             }
         }
--- a/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
+++ b/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
@@ -7,9 +7,6 @@ import CoreData
 /// Should eventually replace objC class `TracksEventPersistenceService`.  Unfortunately
 /// Swift Packages do not support mixed source classes so we'll instead need to compose.
 ///
-/// The naming is slightly different with "Track" instead of "Tracks", but the class is internal for now so it
-/// should be fairly acceptable to do this.
-///
 @objc
 public class TrackEventPersistenceService: NSObject {
     
--- a/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
+++ b/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
@@ -29,7 +29,7 @@ public class TrackEventPersistenceService: NSObject {
 
         managedObjectContext.perform {
             for startIndex in stride(from: 0, to: uuidStrings.count, by: Self.incrementRetryCountBatchSize) {
-                
+
                 let isLastBatch = startIndex + Self.incrementRetryCountBatchSize >= uuidStrings.count
                 let results: [TracksEventCoreData]
                 let count = min(uuidStrings.count - startIndex, Self.incrementRetryCountBatchSize)
--- a/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
+++ b/Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
@@ -9,68 +9,68 @@ import CoreData
 ///
 @objc
 public class TrackEventPersistenceService: NSObject {
-    
+
     private static let incrementRetryCountBatchSize = 500
     private let managedObjectContext: NSManagedObjectContext
-    
+
     @objc
     public init(managedObjectContext: NSManagedObjectContext) {
         self.managedObjectContext = managedObjectContext
-        
+
         super.init()
     }
-    
+
     /// Increments the retry count for the specified events in batches of
     @objc
     public func incrementRetryCountForEvents(_ tracksEvents: [TracksEvent], onComplete completion: ((Error?) -> Void)?) {
         let uuidStrings = tracksEvents.map { event in
             event.uuid.uuidString
         }
-        
+
         for startIndex in stride(from: 0, to: uuidStrings.count, by: Self.incrementRetryCountBatchSize) {
             let isLastBatch = startIndex + Self.incrementRetryCountBatchSize >= uuidStrings.count
-            
+
             managedObjectContext.perform {
                 let results: [TracksEventCoreData]
                 let count = min(uuidStrings.count - startIndex, Self.incrementRetryCountBatchSize)
                 let uuidStringsBatch = Array(uuidStrings[startIndex ..< startIndex + count])
-                
+
                 do {
                     results = try self.findTrackEventCoreData(uuidStrings: uuidStringsBatch)
                 } catch {
                     TracksLogError("Error while finding track events: \(String(describing: error))")
-                    
+
                     if isLastBatch {
                         completion?(error)
                     }
-                    
+
                     return
                 }
-                
+
                 if results.count != count {
                     TracksLogError("Not all provided events were found in the persistence layer. This signals a possible logical error in the tracking, persistence and retry-count-incrementing code.  Please review.")
                 }
-                
+
                 for event in results {
                     event.retryCount = event.retryCount.intValue + 1 as NSNumber
                 }
-                
+
                 self.saveManagedObjectContext()
-                
+
                 if isLastBatch {
                     completion?(nil)
                 }
             }
         }
     }
-    
+
     func findTrackEventCoreData(uuidStrings: [String]) throws -> [TracksEventCoreData] {
         let fetchRequest = NSFetchRequest<TracksEventCoreData>(entityName: "TracksEvent")
         fetchRequest.predicate = NSPredicate(format: "uuid in %@", uuidStrings)
-        
+
         return try fetchRequest.execute()
     }
-    
+
     func saveManagedObjectContext() {
         do {
             try managedObjectContext.save()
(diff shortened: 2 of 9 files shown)


Package.swift
```diff
@@ -60,11 +60,25 @@ let package = Package(
         .target(
             name: "AutomatticTracksEvents",
             dependencies: ["AutomatticTracksModel",
-                           "AutomatticExperiments"],
+                           "AutomatticExperiments",
+                           "AutomatticTracksEventsForSwift"],
```
diegoreymendez, line 64: Since Swift Packages don't support mixed-source, I've created a Swift module to open a path to incrementally migrate module `AutomatticTracksEvents` to Swift (`AutomatticTracksEventsForSwift`).

The idea is that once that migration is complete, it should be possible to simply remove the old module and rename `AutomatticTracksEventsForSwift` > `AutomatticTracksEvents`.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
```diff
@@ -0,0 +1,82 @@
...
+        let uuidStrings = tracksEvents.map { event in
+            event.uuid.uuidString
+        }
+        
+        let batchSize = 500
+        
+        for startIndex in stride(from: 0, to: uuidStrings.count, by: batchSize) {
+            let isLastBatch = startIndex + batchSize >= uuidStrings.count
+            
+            managedObjectContext.perform {
```
diegoreymendez: This is dangerous in my opinion.

It's good in that it allow us to batch the requests to CoreData, but by doing it this way we're effectively allowing other `perform` calls to execute in-between our own `perform` calls, potentially changing the previous behaviour.

I did this because batching sounded like a good idea, but right after finishing with the implementation I wondered if having a single perform block would be better.  We could still keep the batching inside the perform block, although I'm not sure that'd be useful for anything other than memory.

I need some thoughts here.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
```diff
@@ -0,0 +1,82 @@
...
+                    
+                    if isLastBatch {
+                        completion?(error)
+                    }
+                    
+                    return
+                }
+                
+                if results.count != count {
+                    TracksLogError("Not all provided events were found in the persistence layer. This signals a possible logical error in the tracking, persistence and retry-count-incrementing code.  Please review.")
```
diegoreymendez: Partial errors shouldn't halt the process, but we'll probably want them logged just in case.  This type of error shouldn't really happen though, so it may be overkill to even try to log this.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
```diff
@@ -0,0 +1,82 @@
...
+                let count = min(uuidStrings.count - startIndex, batchSize)
+                let uuidStringsBatch = Array(uuidStrings[startIndex ..< startIndex + count])
+                
+                do {
+                    results = try self.findTrackEventCoreData(uuidStrings: uuidStringsBatch)
+                } catch {
+                    TracksLogError("Error while finding track events: \(String(describing: error))")
+                    
+                    if isLastBatch {
+                        completion?(error)
```
diegoreymendez: This is kinda awful in that we'll only see an error if the last call to `findTrackEventCoreData` fails.

IMHO there are two options here:

1. Ignore the partial errors and not return any error ever... just log them.
2. Create a type of error that groups all partial errors and return that instead... although that error could still mean the retry count was successfully incremented for some events.  I'm more prone to rely on option 1 because I don't think these errors are relevant to whoever's calling this method.

Sources/Event Logging/TracksEventPersistenceService.h
```diff
@@ -18,4 +20,11 @@
 
 - (void)incrementRetryCountForEvents:(NSArray *)tracksEvents;
 
+/// Increments the retry count for the specified events, and offers a completion closure with
+/// error handling support.
+/// 
+- (void)incrementRetryCountForEvents:(NSArray *)tracksEvents onComplete:(nullable void(^)(NSError * _Nullable error))completion;
```
diegoreymendez: Only used for testing right now, but this basically offers support to have what we had before (ie: when the increment was run synchronously).

Sources/Event Logging/TracksEventPersistenceService.m
```diff
@@ -20,6 +24,7 @@ - (instancetype)initWithManagedObjectContext:(NSManagedObjectContext *)managedOb
     self = [self init];
     if (self) {
         _managedObjectContext = managedObjectContext;
+        _swiftPersistenceService = [[TrackEventPersistenceService alloc] initWithManagedObjectContext: managedObjectContext];
```
diegoreymendez: Through composition, we can migrate our code to Swift incrementally.

Tests/Tests/TracksEventPersistenceServiceTests.swift
```diff
@@ -0,0 +1,207 @@
...
+        let extraUUID = UUID()
+        let extraEvent = createTestTracksEvent(extraUUID)
+        service.persistTracksEvent(extraEvent)
+        
+        service.incrementRetryCount(forEvents: tracksEvents + [extraEvent]) { error in
+            // Because not finding the events to increase their retry count
+            // are partial errors (ie: they shouldn't stop the overall process), they won't
+            // be reported as an error here, but there should also not be any records for those
+            // events when fetching from core data.
+            XCTAssertNil(error)
```
diegoreymendez: As mentioned previously, I'm starting to doubt the value of having this error.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
```diff
@@ -0,0 +1,81 @@
...
+import AutomatticTracksModel
+import AutomatticTracksModelObjC
+#endif
+import CoreData
+
+/// Should eventually replace objC class `TracksEventPersistenceService`.  Unfortunately
+/// Swift Packages do not support mixed source classes so we'll instead need to compose.
+///
+@objc
+public class TrackEventPersistenceService: NSObject {
```
bjhomer: Is it intentional that this is called **Track**EventPersistenceService, instead of **Tracks**? That does give it a distinct name from the ObjC one, but it seems like it would be really easy to confuse the two. I might recommend something like `TracksEventPesistenceServiceSwift` or something instead.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
bjhomer: The previous code had no room for reporting an error either; maybe we don't even need the `error` parameter in the completion handler here at all? I don't think the caller is ever likely to actually do something about an error here.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
```diff
@@ -0,0 +1,81 @@
...
+                if isLastBatch {
+                    completion?(nil)
+                }
+            }
+        }
+    }
+
+    func findTrackEventCoreData(uuidStrings: [String]) throws -> [TracksEventCoreData] {
+        let fetchRequest = NSFetchRequest<TracksEventCoreData>(entityName: "TracksEvent")
+        fetchRequest.predicate = NSPredicate(format: "uuid in %@", uuidStrings)
```
bjhomer: Since we're going to be modifying each of these objects, it would probably be more efficient to set `fetchRequest.returnsObjectsAsFaults = false`. Otherwise, we'll have to make another trip to the database for each object, which hurts performance. Since we know we're going to need them all anyway, might as well get them all up front.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
```diff
@@ -0,0 +1,81 @@
...
+
+        return try fetchRequest.execute()
```
bjhomer: Huh, I didn't know about `fetchRequest.execute()`. I've always done `context.fetch(fetchRequest)`. This is nice, though, because you don't have to pass in the context. (So long as you're within a `.perform` block, anyway.)

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
bjhomer: Now that this is no longer happening on the main thread, I think it would be fine to keep it all in a single `perform` block. Batching for the sake of memory might be nice, but is probably less critical.

One way to simplify this could be to use `request.fetchBatchSize` to handle the batching automatically. Note, though, that if you're going to use that, you need to interact with the results as an NSArray, not a Swift array. (This is because it actually returns a special subclass of NSArray that fetches things on-demand in batches, but converting it to a Swift array just immediately reads everything, which triggers all the fetching right away and defeats the point of the batching.)

Tests/Tests/CoreData/TracksTestContextManager.swift
```diff
@@ -0,0 +1,60 @@
...
+/// This class duplicates the class found in `TestTracksContextManager.swift`, because there's no support in
+/// packages for mixed-language source files and method-calling.  We could work around this problem
+/// in other ways but since this will eventually move to Swift, I may as well just duplicate it for now and we
+/// can clean-up once we remove the remaining ObjC code.
+///
+class TracksTestContextManager: TracksContextManager {
+
+    private var persistentStoreExists: Bool {
+        get {
+            guard let value = objc_getAssociatedObject(self, &persistentStoreExistsKey) as? Bool else {
```
bjhomer: Why does this need to use associated objects instead of just declaring properties directly?

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
bjhomer: Also, I would recommend adding an index to the data model on `uuid`. That will make this lookup _much_ faster in cases when we have large amounts of events to report.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
```diff
@@ -0,0 +1,81 @@
...
+                self.saveManagedObjectContext()
+
+                if isLastBatch {
+                    completion?(nil)
+                }
+            }
+        }
+    }
+
+    func findTrackEventCoreData(uuidStrings: [String]) throws -> [TracksEventCoreData] {
```
bjhomer: Should this be find**Tracks**EventCoreData?

bjhomer (changes requested):

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Yes.  And as a side note my head keeps telling me this should probably be **findTracksEventsCoreData()** because it's multiple events.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Yes and no.  I had tried using a different name to keep the swift file in the original package, until I learnt that's not possible.

I'll resort to the idea of keeping the same name and having `Swift` in the end.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Maybe since the code for batching is already written I can keep it but move it inside a single perform block.

I think fetching by using an `NSArray` can be hard to understand if someone modifies the code later and doesn't know about that.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: The index idea is spot on, and a great addition to this PR.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Yup!  I was wondering if it could feel a bit "obscure"... but honestly it makes sense that it's using the context of its execution context.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
bjhomer: One note: if you _only_ add an index and nothing else, Core Data won't recognize the data model as needing a migration. So you'll want to also change the value of the "Version Hash Modifier" on `TracksEvent` in the data model. It doesn't matter what you change it _to_; just the fact that it's different will cause the migration mechanism to recognize that something has changed. I usually change it to something describing the last change, such as `added uuid index`.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
bjhomer: Or maybe something like `findCoreDataEvents()`? I don't know that we need the full type name in the function name.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Removed the error from the completion block.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Swapped the perform and loop blocks so that the whole loop happens within a single perform block.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Renamed.

Tests/Tests/CoreData/TracksTestContextManager.swift
diegoreymendez: This class was initially an extension of the ObjC class `TestTracksContextManager` and when I gave up on that plan (due to mixed-language not being supported in Swift Packages), I forgot to change this back into a regular variable.

Good catch!

Changed.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Changed.

Sources/Event Logging (Swift)/TrackEventPersistenceService.swift
diegoreymendez: Added the index and changed the has.  Also no longer allow the events to be returned as faults.

diegoreymendez:
Ready for another look.

Sources/Event Logging (Swift)/TracksEventPersistenceServiceSwift.swift
```diff
@@ -0,0 +1,82 @@
...
+
+                for event in results {
+                    event.retryCount = event.retryCount.intValue + 1 as NSNumber
+                }
+
+                self.saveManagedObjectContext()
+
+                if isLastBatch {
+                    completion?()
+                }
```
bjhomer: We probably don't even need `isLastBatch` anymore; I think we could just put the completion call outside of the `for` loop, right?

diegoreymendez: 🤦  yup

diegoreymendez:
Ready for another look.

diegoreymendez:
Hey @bjhomer if you have a moment this is ready for another look.  Thanks!

bjhomer (approved):
This looks good to me. Thanks!

diegoreymendez:
Thank you!

PR #206, opened 2022-06-14
Update Swift tools version from 5.3 to 5.5

mokagio:
Notice that this introduces an error when running `swift test`:

```
$ swift test

Building for debugging...
/Users/gio/Developer/a8c/tracks-apple-pod/Sources/Model/ObjC/Common/Core Data/TracksContextManager.m:24:24: error: use of undeclared identifier 'SWIFTPM_MODULE_BUNDLE'
    NSBundle *bundle = SWIFTPM_MODULE_BUNDLE;
                       ^
1 error generated.
[0/6] Compiling TracksContextManager.m
error: fatalError
```

This is a known Swift Package Manager issue. See:

- https://github.com/apple/swift-package-manager/issues/4500
- https://forums.swift.org/t/5-3-resources-support-not-working-on-with-swift-test/40381

However, `bundle exec fastlane test`, which uses `xcodebuild` under the hood, works.

Given our workflow is based on Xcode locally and Fastlane in CI, the regression seems acceptable.

The reason I'm keen to upgrade to 5.5 is that it's the required tooling version to integrate Buildkite's Test Analytics. See
https://github.com/buildkite/test-collector-swift/blob/v0.1.0/Package.swift#L1

### DIFF
--- a/.buildkite/check-version-consistency.sh
+++ b/.buildkite/check-version-consistency.sh
@@ -3,6 +3,9 @@
 SRC_FILE=Sources/Model/ObjC/Constants/TracksConstants.m
 PODSPEC_FILE=Automattic-Tracks-iOS.podspec
 
+# Workaround for https://github.com/Automattic/buildkite-ci/issues/79
+gem install bundler
+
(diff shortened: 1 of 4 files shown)


AliSoftware (approved):
Mitigation of the bug with `swift test` (given we use Xcode / `xcodebuild` / `fastlane`, not `swift test`) seems acceptable to me.

PR #208, opened 2022-07-27
Store and confirm what experiments a user is enrolled in

charliescheer:
On Day One iOS right now we are looking to be able to have experiments that are only for new users.  We had been verifying this by using a locally stored and calculated array of the experiments that a user is enrolled in or qualified for.  This has a few issues with it, namely the potential for getting out of sync with what experiments ExPlat thinks a user is enrolled in, and has caused a bug where new user only experiments are being shown to existing users.

This PR is part of the process to fix that bug.  With these changes, when we refresh the experiments I am storing the experiments names that are returned from the server and making that accessible outside of the package to confirm if a user is enrolled for a specific experiment.  We need to store this separately because the server returns the name of the experiment for all enrolled experiments as a key and what the user is assigned to as a value.  If the user is in the control group it returns null, and the existing ab-testing-assignments dictionary ignores the pairs with null as a value.  

I chose to store the values in a separate array rather than change ab-testing-assignments to store the keys even if nothing was returned to avoid creating any potential issues with other apps that are using the experiments values.

With these changes we will be able to more accurately confirm what experiments a user participating in


Sources/Experiments/ExPlat.swift
```diff
@@ -12,6 +12,7 @@ import Cocoa
 
     private let assignmentsKey = "ab-testing-assignments"
     private let ttlDateKey = "ab-testing-ttl-date"
+    private let enrolledKey = "enrolled-experiments"
```
bjhomer: Since this is just being stored straight into UserDefaults with no name spacing, I think it may be good to keep a consistent `ab-testing-` prefix on it. So this would be `ab-testing-enrolled-experiments`.

Sources/Experiments/ExPlat.swift
```diff
@@ -80,6 +88,9 @@ import Cocoa
             var ttlDate = Date()
             ttlDate.addTimeInterval(TimeInterval(assignments.ttl))
             UserDefaults.standard.setValue(ttlDate, forKey: self.ttlDateKey)
+            
+            let enrolledExperimentNames = assignments.variations.map({ String($0.key) })
```
bjhomer: Isn't `$0.key` already a String? What's is the purpose of the `String()` initializer here?

Sources/Experiments/ExPlat.swift
```diff
@@ -100,6 +111,18 @@ import Cocoa
             return .treatment(variation)
         }
     }
+    
+    /// Checks if the experiment name is contained in the enrolled experiments array
+    /// returns false if there is no dictionary or the experiment is not registered
+    ///
+    public func isEnrolled(_ name: String) -> Bool {
```
bjhomer: Could we call this `isEnrolled(in name: String)`?

jleandroperez (approved):
@charliescheer changes look good! please feel free to :shipit: once the Hound warnings are gone sir!

Sources/Experiments/ExPlat.swift
charliescheer: huh... I was sure there was a reason I did that... but yeah, it's a string already.  Removed the type casting

Sources/Experiments/ExPlat.swift
charliescheer: totally

PR #211, opened 2022-08-26
Homogenize `Package.swift` formatting

mokagio:
I've been dipping in and out of the code while trying to understand how to migrate to Sentry version 7.x. As a result, I took some time to indulge in making the `Package.swift` file formatting homogeneous and removing noisy template comments.


jkmassel (approved):
Thanks for taking care of the little things!

PR #214, opened 2022-08-26
Add iOS device accessibility settings to device properties.

selanthiraiyan:
Related to - https://github.com/woocommerce/woocommerce-ios/pull/7576

### Changes
In order to understand the accessibility settings of our users, this PR adds the following properties to the `deviceProperties` dictionary of analytics events.
- preferred content size category to know the font size settings.
- is accessibility category boolean to know whether the font size falls under accessibility category. 

### Known issue

As this PR adds two properties to the `Transformable` `deviceInfo` (a dictionary) property of the core data model, we will face errors when trying to load stores created using previous versions of the core data model. 

This is handled [here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/wcios/track-accessibility-font-setting/Sources/Model/ObjC/Common/Core%20Data/TracksContextManager.m#L51) by deleting the old store and creating a new one using the latest core data model. 

I hope that we have to use a `ValueTransformer` to deal with the `Transformable` `deviceInfo` property of the core data model. I am not sure whether this issue was ignored because,
- A change to the `Transformable` dictionary (`deviceInfo`, `customProperties`, and `userProperties`) type properties is not supposed to happen often. 
- Optimistically, the events are expected to have been already uploaded by the time the user updates the app, and the store gets deleted. 🤞 

Please correct me if I am wrong about this. 

### Testing instructions
Follow the testing instructions from https://github.com/woocommerce/woocommerce-ios/pull/7576

### DIFF
--- a/Sources/Model/ObjC/Common/TracksDeviceInformation.h
+++ b/Sources/Model/ObjC/Common/TracksDeviceInformation.h
@@ -30,7 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
 /// Returns `true` if the preferred reading content size falls under accessibility category.
 ///
 /// - Uses `UIContentSizeCategoryIsAccessibilityCategory` method.
-/// - This will be `false`` for Mac OS.
+/// - This will be `false` for Mac OS.
 ///
 @property (nonatomic, readonly) BOOL isAccessibilityCategory;
 
--- a/Sources/Model/ObjC/Common/TracksDeviceInformation.h
+++ b/Sources/Model/ObjC/Common/TracksDeviceInformation.h
@@ -21,6 +21,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nullable, nonatomic, readonly) NSString *currentNetworkOperator;
 @property (nullable, nonatomic, readonly) NSString *currentNetworkRadioType;
 @property (nonatomic, assign) BOOL isWiFiConnected;
+
+/// Preferred reading content size based on the accessibility setting of the iOS device.
+/// - This will be null for Mac OS.
+///
+@property (nullable, nonatomic, readonly) NSString *preferredContentSizeCategory;
+
 /**
  * Indicates whether the device has an Internet connection.
  */
--- a/Sources/Model/ObjC/Common/TracksDeviceInformation.h
+++ b/Sources/Model/ObjC/Common/TracksDeviceInformation.h
@@ -23,7 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign) BOOL isWiFiConnected;
 
 /// Preferred reading content size based on the accessibility setting of the iOS device.
-/// - This will be null for Mac OS.
+/// - This will be NULL for Mac OS.
 ///
 @property (nullable, nonatomic, readonly) NSString *preferredContentSizeCategory;
 
--- a/Sources/Model/ObjC/Common/TracksDeviceInformation.h
+++ b/Sources/Model/ObjC/Common/TracksDeviceInformation.h
@@ -27,6 +27,13 @@ NS_ASSUME_NONNULL_BEGIN
 ///
 @property (nullable, nonatomic, readonly) NSString *preferredContentSizeCategory;
 
+/// Returns `true` if the preferred reading content size falls under accessibility category.
+///
+/// - Uses `UIContentSizeCategoryIsAccessibilityCategory` method.
+/// - This will be `false`` for Mac OS.
+///
+@property (nonatomic, readonly) BOOL isAccessibilityCategory;
+
 /**
  * Indicates whether the device has an Internet connection.
  */
--- a/Sources/Model/ObjC/Common/TracksDeviceInformation.m
+++ b/Sources/Model/ObjC/Common/TracksDeviceInformation.m
@@ -206,6 +206,28 @@
 }
 #endif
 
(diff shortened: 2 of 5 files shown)


Sources/Model/ObjC/Common/TracksDeviceInformation.m
```diff
@@ -206,6 +206,51 @@ - (UIInterfaceOrientation)statusBarOrientation {
...
+    return preferredContentSizeCategory;
+#else   // Mac
+    return @"Unknown";
+#endif
+}
+
+/// Returns `true` if the preferred reading content size falls under accessibility category.
+///
+/// - Uses `UIContentSizeCategoryIsAccessibilityCategory` method.
+/// - This will be `false`` for Mac OS.
```
jaclync: super nit:
```suggestion
/// - This will be `false` for Mac OS.
```

Sources/Model/ObjC/Common/TracksDeviceInformation.m
```diff
@@ -206,6 +206,51 @@ - (UIInterfaceOrientation)statusBarOrientation {
...
+- (NSString *)preferredContentSizeCategory {
+#if TARGET_OS_IPHONE
+    if ([NSThread isMainThread]) {
+        return UIApplication.sharedIfAvailable.preferredContentSizeCategory;
+    }
+
+    __block NSString *preferredContentSizeCategory;
+
+    dispatch_sync(dispatch_get_main_queue(), ^{
+        preferredContentSizeCategory = UIApplication.sharedIfAvailable.preferredContentSizeCategory;
```
jaclync, line 222: I haven't used Objective-C for a long while 😅 do you know if this pauses the return statement below until this line is called when it's not on the main thread?

Sources/Model/ObjC/Common/TracksDeviceInformation.m
```diff
@@ -206,6 +206,51 @@ - (UIInterfaceOrientation)statusBarOrientation {
...
+    });
+
+    return preferredContentSizeCategory;
+#else   // Mac
+    return @"Unknown";
```
jaclync: this doesn't seem to match the behavior in the comments for this method?

> This will be null for Mac OS.

```suggestion
    return nil; // or NULL in objc?
```

Sources/Model/ObjC/Common/TracksDeviceInformation.h
```diff
@@ -21,6 +21,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/// Preferred reading content size based on the accessibility setting of the iOS device.
+/// - This will be null for Mac OS.
+///
+@property (nullable, nonatomic, readonly) NSString *preferredContentSizeCategory;
+
+/// Returns `true` if the preferred reading content size falls under accessibility category.
+///
+/// - Uses `UIContentSizeCategoryIsAccessibilityCategory` method.
+/// - This will be `false`` for Mac OS.
```
jaclync: super nit:
```suggestion
/// - This will be `false` for Mac OS.
```

jaclync (approved):
LGTM, thanks for taking on this task to help us understand the usage :shipit: 

If I understand the two properties correctly, do they correspond to the two controls in the settings below?



I was trying to verify the new properties in Tracks Live View but couldn't seem to access them probably due to a data backend issue p1661755847008199-slack-C03DKP1JP. The API requests seem to send the expected properties, so we can verify them in Tracks later.

Re Known issue:
I also saw Core Data error messages in Xcode like this:
> CoreData: annotation: 	reason : The model used to open the store is incompatible with the one used to create the store

But I also checked Wormholy that `application_opened` (the first event) is always tracked despite the migration errors. I didn't test the case with unsent events before the migration, but it seems like another [previous PR that adds a new device information property](https://github.com/Automattic/Automattic-Tracks-iOS/pull/182/files) didn't do anything about it either.

Sources/Model/ObjC/Common/TracksDeviceInformation.m
selanthiraiyan: Done in d847e6e9d200776a818e400942cc2eb53e9929fd

Sources/Model/ObjC/Common/TracksDeviceInformation.h
selanthiraiyan: Done in d847e6e9d200776a818e400942cc2eb53e9929fd

Sources/Model/ObjC/Common/TracksDeviceInformation.m
selanthiraiyan: Nice catch. Done in bbb876af3a9c60c4ac671d71889189bc322eb4e7

Sources/Model/ObjC/Common/TracksDeviceInformation.m
selanthiraiyan: It has been a long time for me as well. 😅

> do you know if this pauses the return statement below until this line is called when it's not on the main thread?

Yes, as we dispatching synchronously using `dispatch_sync`, this will wait before executing the following `return` statement.

selanthiraiyan:
>If I understand the two properties correctly, do they correspond to the two controls in the settings below?

Oh, it looks like that. I didn't realise this before. 😄 

I added the `device_info_is_accessibility_category` property with the help of `UIContentSizeCategoryIsAccessibilityCategory` method as it looked like an easy way to filter events.

PR #218, opened 2022-08-30
Isolate call to Sentry `currentHub()`, which will be unavailable in v7.x

mokagio:
The idea is to bypass the code relying in `currentHub()` for the time being, while waiting for Sentry to provide us with some guidance on how to achieve the same result without `currentHub()`.

I'm not in love with this, but it will make the upgrade job more straightforward by not requiring to delete any code that we'd like to keep using in the future.

### DIFF
--- a/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
+++ b/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
@@ -1,0 +1,36 @@
+import Sentry
+
+/// This is an extension on `SentrySDK` to hides the access to the `currentHub()` methods we use in
+/// the codebase.
+///
+/// The reason for this, other than to implement the good old Law of Demeter, is that `currentHub()`
+/// is no longer available in version 7.x. In the future we hope to be able to access this
+/// information anyway, either by Sentry helping us and introducing an alternative way to do what
+/// we need or by us doing some clever private API access. Therefore, we don't want to wholesale
+/// delete the code that used to work in version 6.x, but rather bypass it for the time being.
+extension SentrySDK {
+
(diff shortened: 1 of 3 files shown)


Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
```diff
@@ -0,0 +1,36 @@
...
+    static func currentClient() -> Sentry.Client? {
+        _currentHub()?.getClient()
+    }
+
+    /// Returns the current `SentryHub`.
+    ///
+    /// - Note: Once we'll migrate to version 7.x, this will return `.none` because `currentHub`
+    /// will no longer be available.
+    private static func _currentHub() -> SentryHub? {
+        currentHub()
```
mokagio, line 34: Once we'll update to version 7.x (7.23.0 is the current latest), this will not compile (unless we'll implement more private API accessing) so we'll just need to update it to `.none`.

I think this is an heavy but effective approach to maintain the existing custom code that relies on `currentHub()` to attach the stacktrace to the events we send manually to Sentry so we can hopefully use it again in the future.

mokagio: For the record, I did try to go down the private API path in https://github.com/Automattic/Automattic-Tracks-iOS/commit/5819dd46d05c900e43ac655fa946f69e2bfecb6f.

The code compiles and the demo app runs with it, but then crashes 😢 This is even more the reason why I'm keen on this approach, so that if we decide to allocate time to try hacking our way through this, we'll have a place to start. Of course, I know that Git would have remembered all the code even if we removed all the stacktrace addition implementation that relies on `currentHub()`, but it seems easier to keep the code here while we wait for more information.

crazytonyli (approved):

Issue #217, opened 2022-08-30
Inconsistent macOS unit tests failures in CI


mokagio:
You may have noticed _some_ builds failing the macOS tests step. Emphasis on some, because this doesn't seem to be related to the code but to the agent running the builds.

### The failure

```
[16:45:44]: ▸ Build Succeeded
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticTracksiOS_AutomatticTracksModelObjC' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticRemoteLogging' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticTracksConstantsObjC' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticTracksModelObjC' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticTracksiOS_AutomatticTracksTests' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticEncryptedLogs' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticTracksEventsForSwift' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticCrashLoggingUI' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticTracksModel' from project 'AutomatticTracksiOS')
  | [16:45:44]: ▸ ❌  error: Could not compute signing task inputs - operation was cancelled. (in target 'AutomatticTracks' from project 'AutomatticTracksiOS')
[...]

PR #220, opened 2022-08-31
Add support for Sentry performance monitoring

mokagio:
Version 7.0 introduced performance monitoring. To prove the upgrade was successful, visit the [performance page for the demo app](https://sentry.io/organizations/a8c/performance/summary/?project=5591863&query=&statsPeriod=24h&transaction=ViewController&unselectedSeries=p100%28%29):

image

I also checked that reporting still works. The screenshot below shows events I sent via the Tracks Demo iOS app:

image

_Notice the UTC timestamp in Sentry and the AEST time in my macOS clock._ 

You can also see the individual events here:

- https://sentry.io/organizations/a8c/issues/3532675432/?project=5591863&statsPeriod=24h
- https://sentry.io/organizations/a8c/issues/3530809730/?project=5591863&statsPeriod=24h

### DIFF
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -36,6 +36,6 @@ Pod::Spec.new do |s|
   s.module_name = 'AutomatticTracks'
 
   s.ios.dependency 'UIDeviceIdentifier', '~> 2.0'
-  s.dependency 'Sentry', '~> 7.24.1'
+  s.dependency 'Sentry', '~> 7.25'
   s.dependency 'Sodium', '>= 0.9.1'
 end
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'Automattic-Tracks-iOS'
-  s.version       = '0.12.1-beta.2'
+  s.version       = '0.13.0-beta.1'
 
   s.summary       = 'Simple way to track events in an iOS app with Automattic Tracks internal service'
   s.description   = <<-DESC
--- a/Package.swift
+++ b/Package.swift
@@ -42,11 +42,16 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(name: "Sentry", url: "https://github.com/getsentry/sentry-cocoa", from: "7.24.1"),
-        .package(url: "https://github.com/AliSoftware/OHHTTPStubs", from: "9.0.0"),
-        .package(url: "https://github.com/squarefrog/UIDeviceIdentifier", from: "2.0.0"),
-        .package(name: "OCMock", url: "https://github.com/erikdoe/ocmock", .branch("master")),
+        // Runtime dependencies
+        //
+        // When changing these, make sure to update the matching declaration in
+        // the `podspec` file.
+        .package(name: "Sentry", url: "https://github.com/getsentry/sentry-cocoa", from: "7.25.0"),
         .package(name: "Sodium", url: "https://github.com/jedisct1/swift-sodium", from: "0.9.1"),
+        .package(url: "https://github.com/squarefrog/UIDeviceIdentifier", from: "2.0.0"),
+        // Tests dependencies
+        .package(url: "https://github.com/AliSoftware/OHHTTPStubs", from: "9.0.0"),
+        .package(name: "OCMock", url: "https://github.com/erikdoe/ocmock", .branch("master")),
         .package(name: "BuildkiteTestCollector", url: "https://github.com/buildkite/test-collector-swift", from: "0.3.0"),
     ],
     targets: [
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -56,7 +56,11 @@ public class CrashLogging {
 
             // Performance monitoring options
             options.enableAutoPerformanceTracking = self.dataProvider.enableAutoPerformanceTracking
-            options.tracesSampleRate = NSNumber(value: self.dataProvider.tracesSampleRate)
+            options.tracesSampler = { _ in
+                // To keep our implementation as Sentry agnostic as possible, we don't pass the
+                // input `SamplingContext` down the chain.
+                NSNumber(value: self.dataProvider.tracesSampler())
+            }
             options.enableNetworkTracking = self.dataProvider.enableNetworkTracking
             options.enableFileIOTracking = self.dataProvider.enableFileIOTracking
             options.enableCoreDataTracking = self.dataProvider.enableCoreDataTracking
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -53,6 +53,17 @@ public class CrashLogging {
 
             /// Attach stack traces to non-fatal errors
             options.attachStacktrace = true
+
+            // Performance monitoring options
+            options.enableAutoPerformanceTracking = self.dataProvider.enableAutoPerformanceTracking
+            options.tracesSampleRate = NSNumber(value: self.dataProvider.tracesSampleRate)
+            options.enableNetworkTracking = self.dataProvider.enableNetworkTracking
+            options.enableFileIOTracking = self.dataProvider.enableFileIOTracking
+            options.enableCoreDataTracking = self.dataProvider.enableCoreDataTracking
+            #if os(iOS) || os(tvOS) || targetEnvironment(macCatalyst)
+                options.enableUserInteractionTracing = self.dataProvider.enableUserInteractionTracing
+                options.enableUIViewControllerTracking = self.dataProvider.enableUIViewControllerTracking
+            #endif
         }
 
         Internals.crashLogging = self
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -41,6 +41,11 @@ public extension CrashLoggingDataProvider {
         }
     }
 
+    var tracesSampler: PerformanceTracking.SampleRateGetter {
+        guard case .enabled(let config) = performanceTracking else { return { 0.0 } }
+        return config.sampleRateGetter
+    }
+
     var tracesSampleRate: Double {
         guard case .enabled(let config) = performanceTracking else { return 0.0 }
         return config.sampleRate
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -30,6 +30,10 @@ public extension CrashLoggingDataProvider {
         return false
     }
 
+    /// Performance tracking is disabled by default to avoid accidentally logging what could be a significant number of extra events
+    /// and blow up our budget monitoring.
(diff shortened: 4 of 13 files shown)


Sources/Remote Logging/EventLogging+Sentry.swift
```diff
@@ -22,13 +22,18 @@ extension EventLogging {
             return
         }
 
+        guard let timestamp = event.timestamp else {
```
mokagio, line 25: The `timestamp` property was `Date` in 6.x but is `Date?` in 7.x

Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
```diff
@@ -12,25 +12,30 @@ extension SentrySDK {
...
+    /// Prior to upgrading to Sentry 7.23.0, this method returned:
+    ///
+    /// currentHub()
+    ///
+    /// As of Sentry 7.0.0, `currentHub()` is no longer available. In the future, we hope to be able to use this functionality either via
+    /// private method runtime acces or if Sentry will answer our feature request and make it available. That's the reason why this code
+    /// is here rather than us deleting the entire stack of functionality that relies on it.
     private static func _currentHub() -> SentryHub? {
-        currentHub()
+        .none
```
mokagio: See also the discussion in #218, where this code was introduced.

Package.swift
```diff
@@ -42,11 +42,16 @@ let package = Package(
...
+        // the `podspec` file.
+        .package(name: "Sentry", url: "https://github.com/getsentry/sentry-cocoa", from: "7.23.0"),
+        .package(name: "Sodium", url: "https://github.com/jedisct1/swift-sodium", from: "0.9.1"),
         .package(url: "https://github.com/squarefrog/UIDeviceIdentifier", from: "2.0.0"),
+        // Tests dependencies
+        .package(url: "https://github.com/AliSoftware/OHHTTPStubs", from: "9.0.0"),
         .package(name: "OCMock", url: "https://github.com/erikdoe/ocmock", .branch("master")),
-        .package(name: "Sodium", url: "https://github.com/jedisct1/swift-sodium", from: "0.9.1"),
         .package(name: "BuildkiteTestCollector", url: "https://github.com/buildkite/test-collector-swift", from: "0.3.0"),
     ],
```
mokagio: 😢 that Swift doesn't have a way to specify public vs. internal dependencies.

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -28,4 +29,41 @@ public extension CrashLoggingDataProvider {
...
+        switch performanceTracking {
+        case .enabled: return true
+        case .disabled: return false
+        }
+    }
+
+    var tracesSampleRate: Double {
+        guard case .enabled(let config) = performanceTracking else { return 0.0 }
+        return config.sampleRate
+    }
```
mokagio, line 52: I considered adding something like

```swift
extension PerformanceTracking {

  var configuration: PerformanceTracking.Configuration? { /* return value if .enabled, .none otherwise */ }
}
```

Then simplify these method like:

```swift
performanceTracking.configuration?.tracesSampleRate ?? 0.0
```

I decided against it because I feel like `guard case .enabled...` makes the code more explicit.

Sources/Remote Logging/Crash Logging/PerformanceTracking.swift
```diff
@@ -0,0 +1,42 @@
+/// Defines whether to enable performance tracking, and if so, how to configure it.
+public enum PerformanceTracking {
+
+    case disabled
+    case enabled(Configuration)
```
mokagio: If you've seen the code in `CrashLogging` and `CrashLoggingDataProvider` above, you'll have noticed that the way one can configure performance monitoring in Sentry allows for inconsistent state, or at least _surprising_ state.

`enableAutoPerformanceTracking` is the main switch, if that's `false`, all the other settings are ignored. `tracesSampleRate` also needs to be more than `0.0` or no metric will be captured (fun fact: it defaults to 0.0 in the SDK, despite `enableAutoPerformanceTracking` defaulting to `true`). 

I suppose the Sentry authors decided to adopt this flat structure to preserve Objective-C compatibility. But we don't have to compromise like they do.

This approach, albeit quite verbose, makes undesirable state unrepresentable: For our clients, either performance tracking is `.enabled`, in which case they have a `Configuration` value, or `.disabled` in which case they have no configuration. It's then up to the library's internals to map down to what Sentry expects.

Sources/Remote Logging/Crash Logging/PerformanceTracking.swift
```diff
@@ -0,0 +1,42 @@
...
+        public init(
+            sampleRate: Double,
+            trackCoreData: Bool = true,
+            trackFileIO: Bool = true,
+            trackNetwork: Bool = true,
+            trackUserInteraction: Bool = true,
+            trackViewControllers: Bool = true
+        ) {
+            // Force sample rate to be between 0.0 and 1.0.
+            self.sampleRate = min(max(0.0, sampleRate), 1.0)
```
mokagio: I was tempted to write a custom time for this, but I worry I'm already on the border of over-engineering 😅

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -12,6 +12,7 @@ public protocol CrashLoggingDataProvider {
     var currentUser: TracksUser? { get }
     var additionalUserData: [String: Any] { get }
     var shouldEnableAutomaticSessionTracking: Bool { get }
+    var performanceTracking: PerformanceTracking { get }
 }
```
mokagio, line 16: I was surprised by `CrashLoggingDataProvider` being a `protocol`. I opened #221 to discuss.

mokagio:
@crazytonyli @jkmassel this is ready for review but I'm leaving it a draft so I don't accidentally merge it.

The reason I'm doing this is because I'm keen to hear your thoughts on the version number to use.

My changes a breaking, because of the new property in `CrashLoggingDataProvider` (case in point for it being a `struct` where we can have default values in cases like this, as per #221). If this library was at version 1.x or later, SemVer would require us to ship a new major version. But we're still at 0.x so we have no explicit obligation not to ship breaking changes...

Should we take this as a chance to ship version 1.0? @jkmassel you likely have more background knowledge on this.

cc @jaclync @rachelmcr, you worked on the library recently, too, and might have thoughts on this.

Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
```diff
@@ -1,11 +1,5 @@
 import Sentry
 
-@objc
-protocol SentrySDKInternalMethods {
-    @objc
-    var currentHub: SentryHub { get }
-}
-
```
mokagio, line 8: I moved this at the bottom of the file to make the documentation more prominent.

Automattic-Tracks-iOS.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'Automattic-Tracks-iOS'
-  s.version       = '0.12.1-beta.2'
+  s.version       = '0.13.0-beta.1'
```
mokagio, line 5: The addition of the new, optional, configuration value for performance monitoring grants a minor version bump.

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -28,4 +29,50 @@ public extension CrashLoggingDataProvider {
     var shouldEnableAutomaticSessionTracking: Bool {
         return false
     }
+
+    /// Performance tracking is disabled by default to avoid accidentally logging what could be a significant number of extra events
+    /// and blow up our budget monitoring.
```
mokagio:
```suggestion
    /// and blow up our events budget.
```

Sources/Remote Logging/EventLogging+Sentry.swift
mokagio: The previous implementation (see diff below) used a `Date()` fallback if `event` had no `timestamp`.

I wonder which is more appropriate:

- Handle the case of a missing timestamp (how likely is that?) by using an incorrect one, hence cancelling the timestamp value as a way to determine the correct log file; or
- Bail out because an event without timestamp is inconsistent / unexpected

🤔 

Keen to hear your thoughts. The fix to support falling back to `Date()` is straightforward for me:

```swift
let timestamp: Date
if even.timestamp == nil {
    let now = Date()
    TracksLogDebug("📜 Unable to locate event timestamp. Falling back to current date (\(now))")
    timestamp = now
} else {
    timestamp = timestamp
}
```

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -53,6 +53,21 @@ public class CrashLogging {
...
             /// Attach stack traces to non-fatal errors
             options.attachStacktrace = true
+
+            // Performance monitoring options
+            options.enableAutoPerformanceTracking = self.dataProvider.enableAutoPerformanceTracking
+            options.tracesSampler = { _ in
+                // To keep our implementation as Sentry agnostic as possible, we don't pass the
+                // input `SamplingContext` down the chain.
+                NSNumber(value: self.dataProvider.tracesSampler())
+            }
```
mokagio, line 63: I _could_ move all this closure wrapping logic into the `dataProvider.tracesSampler()` definition. But I think it fit better here where all the code is Sentry-specific.

mokagio:
@crazytonyli  @jkmassel this is officially ready for review. Thanks!

The macOS test may or may not be still failing by the time you read this. More details in #217 .

Sources/Remote Logging/Crash Logging/PerformanceTracking.swift
```diff
@@ -0,0 +1,53 @@
+/// Defines whether to enable performance tracking, and if so, how to configure it.
+public enum PerformanceTracking {
```
crazytonyli, line 2: Have you considered using `Configuration?` to represent enable and disabled performance tracking? By doing this, most of the new getters in `CrashLoggingDataProvider` can be removed.

mokagio: Good point. Extra code is often a liability.

I thought about it at the time but decided to go for the `enum` because it seemed like a more explicit API. 

The value I see in those getters is to test the translation from a `Configuration` value to the Sentry-specific settings.
Of course, one could say, isn't it easier to run:

```swift
guard let configuration = provider.performanceTrackingConfiguration else { return }

// set each Sentry value with a configuration one
```

But to me it felt like those extra test reduced the chance of errors in translating the configuration to only human error in setting "sentry property `a`" with "configuration property `b`".

mokagio: One way I can think of to remove the need for those getters is to add a method like

```swift
func ifEnabled(_ action: (Configuration) -> Void) {
  guard case .enabled(let configuration) = self else { return }
  action(configuration)
}
```

Then we could use it in `CrashLogging` like

```swift
options.enableAutoPerformanceTracking = self.dataProvider.enableAutoPerformanceTracking
self.dataProvider.performanceTracking.ifEnabled { configuration in
    options.enableNetworkTracking = configuration.trackNetwork
    // ...
}
```

crazytonyli: The current code looks good to me. I just thought removing most of the new getters would be a good side effect of using an optional property. I don't think we _need_ to do something extra to remove those new getters.

crazytonyli (approved):

PR #223, opened 2022-09-08
Improves the error logging in `[TracksContextManager persistentStoreCoordinator]`.

diegoreymendez:
Improves the error logging in `[TracksContextManager persistentStoreCoordinator]` to try and detect possible issues with the URLs and prewarming.

This is because we're seeing failures in DO where the file cannot be created.

## Testing:

1. Make these two conditions always succeed by prepending them with `true ||`:

https://github.com/Automattic/Automattic-Tracks-iOS/blob/716bf8f5e06937cdce8801e8be22bf70abf01ce3/Sources/Model/ObjC/Common/Core%20Data/TracksContextManager.m#L49

and 

https://github.com/Automattic/Automattic-Tracks-iOS/blob/716bf8f5e06937cdce8801e8be22bf70abf01ce3/Sources/Model/ObjC/Common/Core%20Data/TracksContextManager.m#L53

2. Add the following line [here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/716bf8f5e06937cdce8801e8be22bf70abf01ce3/Sources/Model/ObjC/Common/Core%20Data/TracksContextManager.m#L54):

```objc
error = [NSError errorWithDomain:"testing" code:1 userInfo:nil];
```

3. Make sure you've set the Sentry's DSN in `Settings.swift` (there's a tracks demo android project) so you can run the demo App.

4. Run the demo app

Make sure you see something like this in your log when it crashes:

```text
Tracks <E>: Unresolved error Error Domain=asd Code=1 "(null)", {
}. Context info {prewarming=0, storeURL=file:///Users/kybos/Library/Developer/CoreSimulator/Devices/056C5AA2-93D7-459C-88FC-8922966C81F0/data/Containers/Data/Application/35DB0537-8EE3-47F1-9510-6F55529B3F84/Library/Application%20Support/com.automattic.TracksDemo/Tracks.sqlite}.
```

### DIFF
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'Automattic-Tracks-iOS'
-  s.version       = '0.12.0'
+  s.version       = '0.12.1-beta.1'
 
   s.summary       = 'Simple way to track events in an iOS app with Automattic Tracks internal service'
   s.description   = <<-DESC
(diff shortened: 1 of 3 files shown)


bjhomer (approved):
Looks good to me!

PR #224, opened 2022-09-08
Updates to sentry 7.24.1

### LINKED ISSUE #181: Blocker for Sentry version 7.x migration
While looking at crash reports as part of the release management duties, I saw a notice from Sentry encouraging to upgrade to their latest version, 7.1.3.

I [opened a WIP PR for it](https://github.com/Automattic/Automattic-Tracks-iOS/pull/180), but, unfortunately, there's a [breaking change between 6.x and 7.x](https://docs.sentry.io/platforms/apple/migration/#migrating-from-6x-to-7x): `SentrySDK` doesn't expose the `currentHub` method, which we use [here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/36db74cceaf98e107e0c713eaae81eb4139c2059/Automattic-Tracks-iOS/Crash%20Logging/CrashLogging.swift#L226) and [here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/36db74cceaf98e107e0c713eaae81eb4139c2059/Automattic-Tracks-iOS/Crash%20Logging/SentryClient+Shim.m#L27) as part of our custom flow to manually log errors with attached stack trace and running a completion block (see https://github.com/getsentry/sentry-cocoa/issues/660).

Before opening an issue on the Sentry repo, I just wanted to double check here whether I may be missing something.

## My current undersanding

We use that `currentHub` method to get the current `SentryHub`. The docs for this type are [sparse](https://docs.sentry.io/platforms/apple/enriching-events/scopes/):

> You can think of the hub as the central point that our SDKs use to route an event to Sentry. When you call init() a hub is created and a client and a blank scope are created on it. That hub is then associated with the current thread and will internally hold a stack of scopes.
>
> The scope will hold useful information that should be sent along with the event.

The docs also imply the average users is not supposed to interact with hubs, hence why they made the getter private, I'm guessing.

> The hub you are unlikely to be interacting with directly unless you are writing an integration or you want to create or destroy scopes.

We use the hub only to get references to the current client and scope. We need an instance of the client because we call a method on it. I'm also pretty sure we need an instance of the scope. "The scope will hold useful information," if we were to init that ourselves because we can't get one from the hub, we'd lose all the current information.

## The way forward

I looped through the autocompletion on `SentrySDK` and `SentryHub` looking for a promising method to supplement what we can't do anymore, but came up short.

I also run the demo app, hacking Tracks so that it would compile, to verify that `Client` responds to the `prepareEvent:withScope:alwaysAttachStacktrace:isCrashEvent:` selector which is the one we built our [shim](https://github.com/Automattic/Automattic-Tracks-iOS/blob/36db74cceaf98e107e0c713eaae81eb4139c2059/Automattic-Tracks-iOS/Crash%20Logging/SentryClient+Shim.m) around. If it didn't, we would have definitely have to go back to the drawing board, but luckily (?) it does.

So, with the info I have at this point, the only way forward I see is opening an issue on the Sentry SDK repo, asking for help, ideally for reintroducing the `currentHub` method.

Here's a draft:

> ## Question
>
> Since [version 7.x removed the `currentHub` and `setCurrentHub` from `SentrySDK`](https://docs.sentry.io/platforms/apple/migration/#migrating-from-6x-to-7x), what's the way to get a reference to the current hub now?
>
> ## Context
> 
> With version 6, we use the hub to get a reference to the current client and current scope in order to attach stack traces to events we log manually.
>
> This relates to https://github.com/getsentry/sentry-cocoa/issues/660.
>
> ## Suggestion
>
> Would it perhaps be possible to reintroduce the `currentHub` getter on `SentrySDK`? Being only a getter method, it should be "harmless" for users to grab a reference and inspect its properties, right? I can see why you wouldn't want users to override the hub, as your docs say "The hub you are unlikely to be interacting with directly unless you are writing an integration or you want to create or destroy scopes," but, for us at least, having the ability to read what's in the hub is crucial.
>
> Thank you for your help

---

What do you think, am I missing something? If not, does my question and suggestion for the Sentry team make sense? Is there anything that I could add to make it easier to understand?

Also, feel free to take a stab at the code in #180 to see if you can get it to work 😄 

Thanks 🙇‍♂️

diegoreymendez:
Updates to Sentry 7.24.1

Resolves https://github.com/Automattic/Automattic-Tracks-iOS/issues/181 (through a hack, though)

## Details:

I posted [a request](https://github.com/getsentry/sentry-cocoa/issues/1451#issuecomment-1240782406) in one of our old issues in the Sentry repo so that we can hopefully avoid having these hacks in-place.

Unfortunately I don't think it's possible to avoid these hacks right now, until we do larger changes on our end.

This PR aims specifically to unblock us and allow us to update to the latest Sentry version.

## Testing:

You'll need to set-up the DSN in `Secrets.swift`.

1. Place breakpoints in the places where I changed the code to see that the selectors don't fail and return reasonable values.
2. Launch the Demo iOS app.
3. Select "Crashes" in the bottom tab bar.
4. Tap on "Send Error and Wait" and inspect the values in the breakpoints you placed to ensure we're getting the right data.

### DIFF
--- a/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
+++ b/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
@@ -30,11 +30,11 @@ extension SentrySDK {
     /// will no longer be available.
     private static func _currentHub() -> SentryHub? {
         let currentHubSelector = #selector(getter: SentrySDKInternalMethods.currentHub)
-        
+
         guard SentrySDK.responds(to: currentHubSelector) else {
             return nil
         }
-        
+
         return SentrySDK.perform(currentHubSelector).takeUnretainedValue() as? SentryHub
     }
 }
--- a/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
+++ b/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
@@ -1,5 +1,11 @@
 import Sentry
 
+@objc
+protocol SentrySDKInternalMethods {
+    @objc
+    var currentHub: SentryHub { get }
+}
+
 /// This is an extension on `SentrySDK` to hides the access to the `currentHub()` methods we use in
 /// the codebase.
 ///
@@ -9,14 +15,6 @@ import Sentry
 /// we need or by us doing some clever private API access. Therefore, we don't want to wholesale
 /// delete the code that used to work in version 6.x, but rather bypass it for the time being.
 extension SentrySDK {
-
-    /// Returns the current `Scope`, or a new one, for the current `SentryHub`.
-    ///
-    /// - Note: Once we'll migrate to version 7.x, this will return `.none` because `currentHub`
-    /// will no longer be available.
-    static func currentScope() -> Scope? {
-        _currentHub()?.getScope()
-    }
 
     /// Returns the `Client` for the current `SentryHub`.
     ///
@@ -31,6 +29,12 @@ extension SentrySDK {
     /// - Note: Once we'll migrate to version 7.x, this will return `.none` because `currentHub`
     /// will no longer be available.
     private static func _currentHub() -> SentryHub? {
-        currentHub()
+        let currentHubSelector = #selector(getter: SentrySDKInternalMethods.currentHub)
+        
+        guard SentrySDK.responds(to: currentHubSelector) else {
+            return nil
+        }
+        
+        return SentrySDK.perform(currentHubSelector).takeUnretainedValue() as? SentryHub
     }
 }
--- a/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
+++ b/Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
@@ -17,17 +17,17 @@ protocol SentrySDKInternalMethods {
 extension SentrySDK {
 
     /// Returns the `Client` for the current `SentryHub`.
+    /// Since this method uses some Sentry internal methods to work, it may cease working entirely and start returning
+    /// `nil` when Sentry is updated.
     ///
-    /// - Note: Once we'll migrate to version 7.x, this will return `.none` because `currentHub`
-    /// will no longer be available.
     static func currentClient() -> Sentry.Client? {
         _currentHub()?.getClient()
     }
 
     /// Returns the current `SentryHub`.
+    /// Since this method uses some Sentry internal methods to work, it may cease working entirely and start returning
+    /// `nil` when Sentry is updated.
     ///
-    /// - Note: Once we'll migrate to version 7.x, this will return `.none` because `currentHub`
-    /// will no longer be available.
     private static func _currentHub() -> SentryHub? {
         let currentHubSelector = #selector(getter: SentrySDKInternalMethods.currentHub)
 
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'Automattic-Tracks-iOS'
-  s.version       = '0.12.1-beta.1'
+  s.version       = '0.12.1-beta.2'
 
   s.summary       = 'Simple way to track events in an iOS app with Automattic Tracks internal service'
   s.description   = <<-DESC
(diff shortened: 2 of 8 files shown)


Sources/Remote Logging/Crash Logging/SentrySDK+CurrentHub.swift
```diff
@@ -31,6 +29,12 @@ extension SentrySDK {
...
     /// will no longer be available.
     private static func _currentHub() -> SentryHub? {
-        currentHub()
+        let currentHubSelector = #selector(getter: SentrySDKInternalMethods.currentHub)
+
+        guard SentrySDK.responds(to: currentHubSelector) else {
+            return nil
+        }
+
+        return SentrySDK.perform(currentHubSelector).takeUnretainedValue() as? SentryHub
```
bjhomer, line 38: Should we update the comment on this method, now that it won't return `.none` anymore?

diegoreymendez: Updated

mokagio (commented):
Great work @diegoreymendez ! I'm so glad you were able to get this to work!

I run it on the Simulator and verified that:

1) the stack trace is present at the app level

image

 2) the stack trace reaches Sentry (example [event)](https://sentry.io/organizations/a8c/issues/3574315072/?project=5591863)

image

I'm a bit unsure why each step is red or why there are danger icons, but I don't think it's a major are of concern because, as observed elsewhere, we have a very limited use case for sending manual events and looking at their stack trace.

Notice also that there are other threads that can be loaded:

image

---

Is this upgrade something you are on a rush to merge? I'm asking because, as you might have seen, I started [working on an upgrade, too](https://github.com/Automattic/Automattic-Tracks-iOS/pull/220).

If you are not in a rush to merge, I would cherry-pick your changes into my branch, which I hope to get to a ready for review place next week. This is also a good time to talk about version numbers, and I [opened a dedicated issue for that](https://github.com/Automattic/Automattic-Tracks-iOS/issues/225).

diegoreymendez:
> I'm a bit unsure why each step is red or why there are danger icons, but I don't think it's a major are of concern because, as observed elsewhere, we have a very limited use case for sending manual events and looking at their stack trace.

I've been seeing these yesterday (outside of my PR), and I think it has to do with Sentry not having the latest symbols for that binary.

> Is this upgrade something you are on a rush to merge? I'm asking because, as you might have seen, I started https://github.com/Automattic/Automattic-Tracks-iOS/pull/220.
>
> If you are not in a rush to merge, I would cherry-pick your changes into my branch, which I hope to get to a ready for review place next week. This is also a good time to talk about version numbers, and I https://github.com/Automattic/Automattic-Tracks-iOS/issues/225.

Is there an advantage in coupling both PRs vs merging this one first?

I'd strongly prefer this be pushed on its own unless there's a reason to want to hold it back.

It will be fairly easy to push iterative improvements on top of these even after they're merged.

mokagio (approved):
> Is there an advantage in coupling both PRs vs merging this one first?
>
> I'd strongly prefer this be pushed on its own unless there's a reason to want to hold it back.

No, nothing major. I've been thinking about SemVer usage recently and from that point of view I would have liked to release the whole Sentry migration in one go. But there's no strong reason for it. Also, it's not like we've been following SemVer practices to the letter in the past so this wouldn't count as a "violation" in my books.

Happy for this to land and for us to keep iterating ♻️ ♻️ ♻️

diegoreymendez:
Thanks @mokagio !

I'm happy to adjust the versioning though so let me know what'd be best!

diegoreymendez:
@mokagio - I'm guessing you're referring to the fact we're doing a big change but not reflecting it through the versioning?

If that's the case I'd be fine with adjusting the versioning and changing the tags I'm using.  I realize it's not ideal after it's published, but I'd be happy to help.

mokagio:
@diegoreymendez the work I've been doing on Sentry v7 is focused at adding support for configuring performance tracking.

My initial implementation introduced a breaking change (but then @leandroalonso [suggested](https://github.com/Automattic/Automattic-Tracks-iOS/issues/225#issuecomment-1241914832) a [non-breaking alternative](https://github.com/Automattic/Automattic-Tracks-iOS/pull/220/commits/ac3423d557ec7d196a3d89989ae282af5b9db4fc)). 

When I left the comment, I was still thinking in terms of shipping a breaking change and I would have liked to reflect it in the version number. Under the new circumstances I think:

- The change in this PR is appropriate as a patch update since it "only" updates and internal (sort of) library
- The upcoming change to add performance tracking support can be a minor version bump: new feature, but not breaking

_Thank you for offering to help with updating the release version anyway. I really appreciate it._

diegoreymendez:
Thanks @mokagio!  That makes sense, and I agree that since most of what we're doing is "invisible" to the Apps it may be fine to go for minor version increases.

PR #229, opened 2022-09-28
Update ABTesting Variation representation.

### LINKED ISSUE #226: Change `Variation.treatment` to take a non-optional string
Currently, [`Variation.treatment` has an associated value of `String?`](https://github.com/Automattic/Automattic-Tracks-iOS/blob/79fc1521ce8c03162ad84793e20cdf95f34421fc/Sources/Experiments/ABTesting.swift#L3-L6). It looks like this:

```swift
public enum Variation: Equatable {
    case control
    case treatment(String?)
}
```

I'm proposing that it should be this instead:

```swift
    case treatment(String)
```

`.treatment(nil)` is currently used to represent the default treatment case in a A/B test. You can [see here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/79fc1521ce8c03162ad84793e20cdf95f34421fc/Sources/Experiments/ExPlat.swift#L107-L108) that when the ExPlat server sends back the string `"treatment"`, we represent that as `.treatment(nil)`.

ExPlat now has added beta support for `A/B/n` experiments, where you can have multiple named variations in addition to the control. This API supports that just fine; you might get `.treatment("green_button")` or `.treatment("blue_button")`, and we should definitely continue supporting that.

**What I'd like to propose** is that we stop mapping `"treatment"` to `nil`. Instead, the default A/B variant would simply be `.treatment("treatment")`. I would argue that doing so is clearer, and no more difficult to work with.

I've checked the other Automattic apps I have access to, and I found [only one case](https://github.com/woocommerce/woocommerce-ios/blob/db28955c8fc78e7ed9c7c9efbec56637987693c5/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/ProductFormActionsFactory.swift#L80) so far where this would be a breaking change. That case could easily be updated to check for `.treatment("treatment")` instead. If we'd like to simplify checking for "is this any kind of treatment at all", we could add the following computed property to `Variation`:

```swift
extension Variation {
  public var isTreatment: Bool {
    switch self {
    case .control: return false
    case .treatment: return true
    }
  }
}
```

bjhomer:
Most AB tests run with just `control` and `treatment` variations. These were curiously represented in the Tracks code as `.control` and `.treatment(nil)`, respectively. That `nil` is quite confusing for such a common case.

This PR updates the representation of the common case to simply be `.treatment`. Custom cases are now represented with a separate enum value.

Treatment name from server | Old representation | New representation
---|---|---
`"control"` | `.control` | `.control`
`"treatment"` | `.treatment(nil)` | `.treatment`
`"purple_button"` | `.treatment("purple_button")` | `.customTreatment(name: "purple_button")`

Fixes #226

### DIFF
--- a/Sources/Experiments/ExPlat.swift
+++ b/Sources/Experiments/ExPlat.swift
@@ -105,9 +105,9 @@ import Cocoa
         case "control":
             return .control
         case "treatment":
-            return .treatment(nil)
-        default:
-            return .treatment(variation)
+            return .treatment
+        case let name:
+            return .customTreatment(name: name)
         }
     }
 
--- a/Sources/Experiments/ABTesting.swift
+++ b/Sources/Experiments/ABTesting.swift
@@ -2,7 +2,8 @@ import Foundation
 
 public enum Variation: Equatable {
     case control
-    case treatment(String?)
+    case treatment
+    case customTreatment(name: String)
 }
 
 /// A protocol that defines a A/B Testing provider
--- a/Tests/Tests/ABTesting/ExPlatTests.swift
+++ b/Tests/Tests/ABTesting/ExPlatTests.swift
@@ -22,8 +22,8 @@ class ExPlatTests: XCTestCase {
 
         abTesting.refresh {
             XCTAssertEqual(abTesting.experiment("experiment"), .control)
-            XCTAssertEqual(abTesting.experiment("another_experiment"), .treatment(nil))
-            XCTAssertEqual(abTesting.experiment("experiment_multiple_variation"), .treatment("another_treatment"))
+            XCTAssertEqual(abTesting.experiment("another_experiment"), .treatment)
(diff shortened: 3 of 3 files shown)


Sources/Experiments/ExPlat.swift
```diff
@@ -105,9 +105,9 @@ import Cocoa
         case "control":
             return .control
         case "treatment":
-            return .treatment(nil)
-        default:
-            return .treatment(variation)
+            return .treatment
+        case let name:
+            return .customTreatment(name: name)
```
mokagio, line 110: Nitpick that might be out of scope for this PR, but this code seems like it would fit nicely in a custom initializer for `Variation`:

```
return Variation(from: variation)
```

mokagio (approved):
Looking good 👍

PR #231, opened 2022-10-17
Dispatch an UI update event in the macOS demo app from the main thread

mokagio:
Xcode showed a Main Thread Sanitizer warning about this:

image

If you run the app on this branch, you'll see no warning.

_I noticed this while experimenting with Sentry's `flush` in #232._


mokagio (commented):
There might be a more appropriate place where to ensure the threading is correct. If you can point it out, I'll happily update.

However, given this is a demo app for dev/testing purposes, I didn't sweat the details.

crazytonyli (approved):

PR #232, opened 2022-10-17
Use Sentry `flush` to send events immediately

mokagio:
This PR removes our custom logic to implement the "send all enqueue events and wait" functionality in favor of using the new Sentry's native `flush` method.

One big disadvantage of this approach is that we lose the precise callback management, because [`flush` is blocking for as long as it takes to send all events or for a given timeout to run out](https://github.com/getsentry/sentry-cocoa/issues/1013#issuecomment-1243634969).

I used the demo app to verify our custom implementation works and the events it sends reach Sentry. In the screenshot below, the top two events have been sent from this branch. Also notice the third and fourth events have "(no value)" for release and environment. That's a bug introduced at some point when migrating to version Sentry version 7. I haven't looked into it yet.

image

I'm of two minds on whether to go for using `flush` under the hood. I really dislike how `flush` doesn't provide a callback out of the box, and how it results in a less precise UI state handling on our end. At the same time, relying on Sentry's native code instead of implementing our own logic accessing their private API simplifies and de-risks the library.

I'd like other folks input on this, but I'm tempted to adopt this implementation on the basis that `logErrorAndWait`, `logErrorImmediately`, and `logErrorsImmediately` are used very sparingly by the library's clients. Namely, [in one occasion in Jetpack/WordPress](https://github.com/wordpress-mobile/WordPress-iOS/search?q=logErrorAndWait) and in two occasions in WooCommerce](https://github.com/woocommerce/woocommerce-ios/search?q=logFatalErrorAndExit).


Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -23,6 +23,9 @@ public class CrashLogging {
     /// crash log events will only be sent in Release builds.
     public static let forceCrashLoggingKey = "force-crash-logging"
 
+    // TODO: This could be made a configurable instance property, with default value in the `init`
+    public static let eventsFlushingTimeout: TimeInterval = 15
```
mokagio: Waiting to act upon this till we decided whether to go down the Sentry `flush` route.

Sources/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -97,14 +97,11 @@ extension CrashLoggingView {
...
-                switch result {
-                    case .success:
-                        sendErrorAndWaitStatus = .success
-                    case .failure(let err):
-                        sendingError = err
-                        sendErrorAndWaitStatus = .error
-                }
+            try crashLogging.logErrorImmediately(
+                error,
+                userInfo: ["custom-userInfo-key": "custom-userInfo-value"]
```
mokagio: I realized we never "tested" sending custom information, so I added it. Here's the result:

image

Sources/UI/Crash Logging/CrashLoggingView.swift
```diff
@@ -97,14 +97,11 @@ extension CrashLoggingView {
...
+            ) {
+                sendErrorAndWaitStatus = .success
```
mokagio: If we decide to adopt `flush`, it might be appropriate to update this `enum` to remove the error-tracking state, since there's no longer any error information with `flush`.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -168,54 +171,24 @@ public extension CrashLogging {
     }
 
     /// Sends an `Event` to Sentry and triggers a callback on completion
-    func logErrorImmediately(_ error: Error, userInfo: [String: Any]? = nil, level: SentryLevel = .error, callback: @escaping (Result<Bool, Error>) -> Void) throws {
+    func logErrorImmediately(_ error: Error, userInfo: [String: Any]? = nil, level: SentryLevel = .error, callback: @escaping () -> Void) throws {
```
mokagio: Surprisingly, I'm not getting any errors or warnings from the compiler because of these methods still having `throws` despite there not being any genuine `try` call in the chain.

If we commit to `flush`, we should remove `throws` as well.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -168,54 +171,24 @@ public extension CrashLogging {
...
-        URLSession.shared.dataTask(with: request) { (responseData, urlResponse, error) in
-            if let error = error {
-                callback(.failure(error))
-                return
-            }
-
-            let didSucceed = 200...299 ~= (urlResponse as! HTTPURLResponse).statusCode
-            callback(.success(didSucceed))
-        }.resume()
+        SentrySDK.flush(timeout: CrashLogging.eventsFlushingTimeout)
```
crazytonyli: Putting this into a background queue so that we don't block `logErrorsImmediately` callers?

crazytonyli: Or adding a `wait: Bool` argument, which'd make the `logErrorAndWait` simpler to implement.

crazytonyli (changes requested):

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -187,8 +187,12 @@ public extension CrashLogging {
...
             }
         }
-        SentrySDK.flush(timeout: CrashLogging.eventsFlushingTimeout)
-        callback()
+
+        let queue = DispatchQueue(label: "com.automattic.tracks.flushSentry", qos: .background)
+        queue.async {
+            SentrySDK.flush(timeout: CrashLogging.eventsFlushingTimeout)
+            callback()
+        }
```
mokagio, line 221: @crazytonyli what do you think of this as an implementation of your suggestion to make the call in a background queue?

mokagio: In regards to your suggestion to [add a `wait: Bool`](https://github.com/Automattic/Automattic-Tracks-iOS/pull/232#discussion_r999047433) parameter, are you thinking of something along the lines [of WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/feb1cc189cf70c050f8bc5aa74630d4b9dff34a2/WordPress/Classes/Utility/ContainerContextFactory.swift#L23-L35) where we use a `Bool` parameter to decide whether to call `context.performAndWait(_:)` or `context.perform(_:)`?

In this, I'm thinking if `wait == true` we wouldn't run the code in `queue.async`.

crazytonyli: Yes, so that we don't need to use semaphore like the original implementation 😄

mokagio: 😕 The new diff makes this conversation odd because the code it referred to has moved. Anyways... @crazytonyli I had an initial go at adding a `wait` parameter. 

I did this in the form of a `private` func for the other methods to call.

```swift
private func logErrorsImmediately(
        _ errors: [Error],
        userInfo: [String: Any]? = nil,
        level: SentryLevel = .error,
        andWait wait: Bool,
        callback: @escaping () -> Void
    ) {
```

I didn't modify the original method signatures to avoid a breaking change, but also because I wasn't sure how it would have felt to have a `andWait:,` together with a `callback:`. If `wait == true`, then there's no need for a callback because the code that would run in there essentially runs once the method has finished anyway.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
crazytonyli: Any plan for addressing this TODO now that this PR is no longer draft?

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -225,64 +186,40 @@ public extension CrashLogging {
...
-            "sentry_timesetamp=\(Date().timeIntervalSince1970)",
-            "sentry_key=\(user)",
-        ]
-
-        if let password = url.password {
-            data.append("sentry_secret=\(password)")
+        if wait {
+            flushEventThenCallCallback()
+        } else {
+            let queue = DispatchQueue(label: "com.automattic.tracks.flushSentry", qos: .background)
```
crazytonyli: Any reason why using a custom queue instead of `DispatchQueue.global`? Also, any reason why choose `.background` QoS, instead of the default one? (I almost always use the default QoS since I've heard scary things about QoS but don't really understand it 😸 )

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -168,54 +171,12 @@ public extension CrashLogging {
...
-        URLSession.shared.dataTask(with: request) { (responseData, urlResponse, error) in
-            if let error = error {
-                callback(.failure(error))
-                return
-            }
-
-            let didSucceed = 200...299 ~= (urlResponse as! HTTPURLResponse).statusCode
-            callback(.success(didSucceed))
-        }.resume()
+    func logErrorsImmediately(_ errors: [Error], userInfo: [String: Any]? = nil, level: SentryLevel = .error, callback: @escaping () -> Void) throws {
```
crazytonyli: I had a quick look, it appears we can't get the result of `flush` call, is that correct? I think it would be good to at least have a "success or not" result, if we can't get a concrete error. But I guess there isn't anything we can do if Sentry doesn't expose such API.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
mokagio: Good point. Done in 30a873ed4e02b54b77c8e87b6e3fa693da41d250

Sources/Remote Logging/Crash Logging/CrashLogging.swift
mokagio: Correct. See the details given [here](https://github.com/getsentry/sentry-cocoa/issues/1013). It's quite the let down. There must be a reason for the lack of callback in their [unified API](https://develop.sentry.dev/sdk/unified-api) rationale, but to be honest I haven't read that documentation.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
mokagio: I followed the suggestion from one of the Sentry engineers from [here](https://github.com/getsentry/sentry-cocoa/issues/1013#issuecomment-1244958158).

I trust your advice on this matter, though. I'll update.

mokagio: Done in [Automattic/Automattic-Tracks-iOS@`8a0c84b` (#232)](https://github.com/Automattic/Automattic-Tracks-iOS/pull/232/commits/8a0c84b201c7b919aa6a2b8b752e0dcb3269c0a9)

crazytonyli (approved):

PR #233, opened 2022-10-19
Add missing icons to demo app to address Xcode warning

mokagio:
_Stumbled on this while working on #232._

To verify this change, checkout `trunk` and run the tests on the iPhone target. You'll see two warnings about missing icons. Repeat on this branch, and the warnings will be gone.

image

For the record, I created the missing images by selecting the biggest existing one and resizing it to 1024x1024px first and 167x167px later. Given this is a demo app, app icon quality is not something we care about and a crude approach like the one I used is good enough for us.


crazytonyli:
Cute icons 😃

crazytonyli (approved):

PR #236, opened 2022-11-08
Update `CrashLoggingView` after adopting Sentry `flush`, and remove dead `throws`

mokagio:
We no longer have error handling because of Sentry's `flush`, so having `success` and `error` states would be misleading. Likewise, none of the leftover code `throw` errors and there is no point in requiring callers to add `do catch` to use this code.

This is leftover work from #232. That PR has already been approved, so I created a new one to avoid noise.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ _None._
 ### Breaking Changes
 
 - `logErrorImmediately` and `logErrorsImmediately` no longer have a `Result` parameter in their callback [#232]
+- `logErrorImmediately` and `logErrorsImmediately` no longer `throws` [#236]
 
 ### New Features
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ _None._
 ### Breaking Changes
 
 - `logErrorImmediately` and `logErrorsImmediately` no longer have a `Result` parameter in their callback [#232]
+- `logErrorImmediately` and `logErrorsImmediately` no longer `throws` [#236]
```
mokagio, line 38: The change in the `SendErrorAndWaitStatus` `enum` is not mentioned because the type is `internal`.

crazytonyli (approved):

Issue #225, opened 2022-09-09
Time to cut version 1.0.0?


mokagio:
While working on updating Sentry to version 7 and adding support for performance monitoring, I introduced a breaking change in the form of [a new required property in `CrashLoggingDataProvider`](https://github.com/Automattic/Automattic-Tracks-iOS/pull/220/files#diff-c5b462d3cd9656fc4b2a17877aa1cfa1f1252159a8f16dfb9a8b138c3f2865aeR15).

This is no big deal. Tracks is an internal library with only a few clients and we control the update process for all of them. The library is also at version `0.x`, and according to SemVer one can expect breaking changes when in this pre release state.

Still, it made me wonder if there's any reason why we haven't cut a `1.0.0` version of this library yet and whether it's a good time to do it.

What would it take to reach `1.0.0`? Is it just a matter of bumping the major version value and calling it a day, or are there API signatures that we're unhappy with and should update first?

mokagio:
cc @bjhomer as you've done lots of work with the library when adding SPM support.

leandroalonso:
@mokagio one way to overcome this without affecting the current clients is to just provide a default value in a protocol extension — maybe if that's going to be adopted only by some apps might make sense. But the new property it's a minor change so I'm not so worried.

What worries me a bit is how Tracks is becoming more and more coupled to Sentry. I don't have enough context as to the decisions of why, but if we decide to change the crash tool in the future we're in for quite an adventure. But I'm deviating from your original question, so for me, it's ok to cut a `1.0.0`.

bjhomer:
> are there API signatures that we're unhappy with and should update first?

I've actually been wanting to propose one API change related to A/B test variations. [I filed it here](https://github.com/Automattic/Automattic-Tracks-iOS/issues/226).

mokagio:
@leandroalonso:

> What worries me a bit is how Tracks is becoming more and more coupled to Sentry. I don't have enough context as to the decisions of why, but if we decide to change the crash tool in the future we're in for quite an adventure. But I'm deviating from your original question, so for me, it's ok to cut a 1.0.0.

That's a good concern. We're trying to keep the implementation agnostic but of course with Sentry being our only crash reporting client, the our API shape kinda is somehow molded on theirs.

@bjhomer awesome, thanks for pointing that out. I don't see any rush in tagging a 1.0.0, so we should definitely iron out any possible breaking change before moving forward with it 👍

mokagio:
[...]

PR #239, opened 2022-12-06
Print a log message if user blocked events sending

mokagio:
While [investigating the `flush` behavior in WooCommerce](https://github.com/Automattic/Automattic-Tracks-iOS/issues/225#issuecomment-1324924475), it took me more than I care to admit to realize I wasn't seeing events because my Simulator was simply not sending them.

This additional log should help realize that sooner in the future.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ _None._
 
 ### Internal Changes
 
-_None._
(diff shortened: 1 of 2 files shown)


crazytonyli (approved):
😆 I can see how I may waste a long time on this issue too

PR #244, opened 2023-01-23
Remove Swift Package Manager workaround for Xcode 12

bjhomer:
We had an extra target in Package.swift to fix a build issue when using Xcode 12. We no longer need the workaround since we're not building with Xcode 12 anymore, so I'm removing it.

---


mokagio (approved):
Builds locally and in CI. Dropping support for Xcode 12 is a no brainer.

Thanks @bjhomer

PR #247, opened 2023-02-02
Experiment Variation - Return optional instead of `control` 

selanthiraiyan:
Related to - https://github.com/woocommerce/woocommerce-ios/pull/8806

## Changes
- `ExPlat` returns optional instead of assuming `control` as experiment variation. This lets the client know that there is no variation for an experiment.
- Make `Variation` confirm to codable.

## Why
In WooCommerce iOS, we will cache the variations for the logged-out experiments. We are doing this to try to remove the high crossovers in Abacus that we face for logged-out experiments.

Choosing this solution after exploring other options. More background in https://github.com/woocommerce/woocommerce-ios/pull/8788

## Testing

- CI should pass successfully.
- Test https://github.com/woocommerce/woocommerce-ios/pull/8806

---

### DIFF
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'Automattic-Tracks-iOS'
-  s.version       = '1.0.0'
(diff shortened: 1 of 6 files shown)


itsmeichigo (approved):
LGTM :shipit:

PR #245, opened 2023-01-23
Upgrade to Sentry SDK 8.3, and expose profiling as an option.

bjhomer:
See https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md for details on what's new in Sentry 8.0

Profiling provides [code-level performance profiling](https://docs.sentry.io/product/profiling/), allowing developers to find out exactly which lines of code are slow. At this point, Sentry still calls profiling "Experimental". We'd like to have the option to try it out.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #253, opened 2023-04-25
Groundwork to support ExPlat in Jetpack

salimbraksa:
Ref p1682352977248619-slack-C04M1NX8YCR

### Related PRs
This PR is blocking:
- https://github.com/wordpress-mobile/WordPress-iOS/pull/20607
- https://github.com/wordpress-mobile/WordPress-iOS/pull/20590

#### The Problem

ExPlat works in WordPress, but it doesn't in Jetpack. This is because of the following endpoint that only supports WordPress:

```swift
var assignmentsEndpoint: String {
    return "https://public-api.wordpress.com/wpcom/v2/experiments/0.1.0/assignments/\(platform)"
}
```

The `platform` variable is substituted with `wpios` when running `WordPress` and `jpios` when running Jetpack. The problem is that the Backend doesn't support the `.../assignments/jpios` endpoint.

#### Workaround

The Workaround that's already implemented in Android ( see p1682483378071949/1682352977.248619-slack-C04M1NX8YCR ) is to call `.../assignments/wpios` endpoint even on Jetpack.

---

### DIFF
--- a/Sources/Event Logging/TracksService.h
+++ b/Sources/Event Logging/TracksService.h
@@ -12,7 +12,7 @@ extern NSString *const TrackServiceDidSendQueuedEventsNotification;
 @property (nonatomic, strong) TracksServiceRemote *remote;
 
 /*! Used to configure the ExPlat `platform` property. If this property is `nil`, the `eventNamePrefix` property is used instead. Default is `nil`. */
-@property (nonatomic, copy, nullable) NSString *platform;
+@property (nonatomic, copy) NSString *platform;
 
 @property (nonatomic, copy) NSString *eventNamePrefix;
 @property (nonatomic, copy) NSString *anonymousUserTypeKey;
--- a/Sources/Event Logging/TracksService.h
+++ b/Sources/Event Logging/TracksService.h
@@ -11,6 +11,9 @@ extern NSString *const TrackServiceDidSendQueuedEventsNotification;
 @property (nonatomic, strong) TracksEventService *tracksEventService;
 @property (nonatomic, strong) TracksServiceRemote *remote;
 
+/*! Used to configure the ExPlat `platform` property. If this property is `nil`, the `eventNamePrefix` property is used instead. Default is `nil`. */
+@property (nonatomic, copy, nullable) NSString *platform;
+
 @property (nonatomic, copy) NSString *eventNamePrefix;
 @property (nonatomic, copy) NSString *anonymousUserTypeKey;
 @property (nonatomic, copy) NSString *authenticatedUserTypeKey;
--- a/Sources/Event Logging/TracksService.m
+++ b/Sources/Event Logging/TracksService.m
@@ -83,6 +83,7 @@ NSString *const USER_ID_ANON = @"anonId";
 {
     self = [super init];
     if (self) {
+        _platform = nil;
         _eventNamePrefix = @"wpios";
         _anonymousUserTypeKey = TracksUserTypeAnonymous;
         _authenticatedUserTypeKey = TracksUserTypeAuthenticated;
@@ -229,7 +230,8 @@ NSString *const USER_ID_ANON = @"anonId";
     self.token = token;
 
     #if TARGET_OS_IPHONE
-    [ExPlat configureWithPlatform:_eventNamePrefix oAuthToken:token userAgent:self.userAgent anonId:anonymousID];
+    NSString *platform = _platform ? _platform : _eventNamePrefix;
+    [ExPlat configureWithPlatform:platform oAuthToken:token userAgent:self.userAgent anonId:anonymousID];
     #endif
 }
 
@@ -243,7 +245,8 @@ NSString *const USER_ID_ANON = @"anonId";
     self.token = nil;
 
     #if TARGET_OS_IPHONE
-    [ExPlat configureWithPlatform:_eventNamePrefix oAuthToken:nil userAgent:self.userAgent anonId:anonymousID];
+    NSString *platform = _platform ? _platform : _eventNamePrefix;
+    [ExPlat configureWithPlatform:platform oAuthToken:nil userAgent:self.userAgent anonId:anonymousID];
     #endif
 }
 
--- a/Sources/Event Logging/TracksService.m
+++ b/Sources/Event Logging/TracksService.m
@@ -230,8 +230,7 @@ NSString *const USER_ID_ANON = @"anonId";
     self.token = token;
 
     #if TARGET_OS_IPHONE
-    NSString *platform = _platform ? _platform : _eventNamePrefix;
-    [ExPlat configureWithPlatform:platform oAuthToken:token userAgent:self.userAgent anonId:anonymousID];
+    [ExPlat configureWithPlatform:[self platformForExPlat] oAuthToken:token userAgent:self.userAgent anonId:anonymousID];
     #endif
 }
 
@@ -245,8 +244,7 @@ NSString *const USER_ID_ANON = @"anonId";
     self.token = nil;
 
     #if TARGET_OS_IPHONE
-    NSString *platform = _platform ? _platform : _eventNamePrefix;
-    [ExPlat configureWithPlatform:platform oAuthToken:nil userAgent:self.userAgent anonId:anonymousID];
+    [ExPlat configureWithPlatform:[self platformForExPlat] oAuthToken:nil userAgent:self.userAgent anonId:anonymousID];
     #endif
 }
 
@@ -260,6 +258,11 @@ NSString *const USER_ID_ANON = @"anonId";
     } else {
         [self.timer invalidate];
     }
+}
+
+- (NSString * _Nullable)platformForExPlat
+{
+    return _platform ? _platform : _eventNamePrefix;
 }
 
 #pragma mark - Reachability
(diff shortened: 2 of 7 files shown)


Sources/Event Logging/TracksService.h
```diff
@@ -11,6 +11,9 @@ extern NSString *const TrackServiceDidSendQueuedEventsNotification;
 @property (nonatomic, strong) TracksEventService *tracksEventService;
 @property (nonatomic, strong) TracksServiceRemote *remote;
 
+/*! Used to configure the ExPlat `platform` property. If this property is `nil`, the `eventNamePrefix` property is used instead. Default is `nil`. */
+@property (nonatomic, copy, nullable) NSString *platform;
```
salimbraksa: The `WordPress-iOS` repo will set this to `wpios` for both Jetpack and WordPress targets.

Sources/Event Logging/TracksService.m
```diff
@@ -229,7 +230,8 @@ - (void)switchToAuthenticatedUserWithUsername:(NSString *)username userID:(NSStr
     self.token = token;
 
     #if TARGET_OS_IPHONE
-    [ExPlat configureWithPlatform:_eventNamePrefix oAuthToken:token userAgent:self.userAgent anonId:anonymousID];
+    NSString *platform = _platform ? _platform : _eventNamePrefix;
+    [ExPlat configureWithPlatform:platform oAuthToken:token userAgent:self.userAgent anonId:anonymousID];
```
salimbraksa: To avoid breaking other repositories, this property is `nil` by default, and `eventNamePrefix` is used instead.

salimbraksa:
Hey @mokagio, any idea why [buildkite/tracks-ios/validating-podspec](https://buildkite.com/automattic/tracks-ios/builds/338#0187bde2-2c66-4521-91c7-1bb8195675bb) check is failing?

crazytonyli:
@salimbraksa If you use the "Jump to end" feature on buildkite, you'll see [the error message](https://buildkite.com/automattic/tracks-ios/builds/338#0187bff9-51e5-4174-9d60-ef3084f49bbc/6-4382):
> [!] Automattic-Tracks-iOS did not pass validation, due to 72 warnings (but you can use `--allow-warnings` to ignore them). You can use the `--no-clean` option to inspect any issue.

Which is probably introduced by [the new platform property](https://github.com/Automattic/Automattic-Tracks-iOS/pull/253#discussion_r1177909908).

TracksDemo/TracksDemo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/getsentry/sentry-cocoa",
         "state": {
           "branch": null,
-          "revision": "dc00a3673e3d313e41326bd53802f757defd0eee",
-          "version": "7.28.0"
+          "revision": "2479b6f7ff69b66bcdea82184d097667e63828ed",
+          "version": "8.5.0"
```
mokagio, line 37: @salimbraksa I noticed you reverted this change in d23173c215b0727861800a5fbbbaa25ca50c14da, but your Xcode must have snuck it in again. 

FWIW, I think it's appropriate to keep it, even though it looks out of place in this PR. It should have been part of https://github.com/Automattic/Automattic-Tracks-iOS/pull/245, which updated Sentry in the main `Package.resolved` file.

https://github.com/Automattic/Automattic-Tracks-iOS/blob/5e42ccaee51d6cd7c9e0f64f2a939a6ee32ba44b/Package.resolved#L28

This file is irrelevant for consumers of the library, and only used by Xcode to manage the versions in the demo app workspace.

Sources/Event Logging/TracksService.m
```diff
@@ -243,7 +245,8 @@ - (void)switchToAnonymousUserWithAnonymousID:(NSString *)anonymousID
     self.token = nil;
 
     #if TARGET_OS_IPHONE
-    [ExPlat configureWithPlatform:_eventNamePrefix oAuthToken:nil userAgent:self.userAgent anonId:anonymousID];
+    NSString *platform = _platform ? _platform : _eventNamePrefix;
```
mokagio: This `platform` conditional assignment logic is duplicated at line 233 above.

Have you considered extracting it in an helper method? `[self platformForExPlat]`?

Alternatively, since it seem like `[ExPlat configureWithPlatform:oAuthToken:userAgent:anonId:]` is the only consumer of this computed `platform` value, what do you think of moving the computation in there?

mokagio: This workaround seems critical for the correct attribution of events and experiments but there are no tests for it.

salimbraksa: >This workaround seems critical for the correct attribution of events and experiments but there are no tests for it.

Resolved in 2f694f46df1052427097dc025a366fdf73e7147d

Also, I want to clarify that the addition of the `platform`  property only impacts **ExPlat**, Track events are still using the `eventNamePrefix` property.

salimbraksa: > Have you considered extracting it in an helper method? [self platformForExPlat]?

Resolved in d15c2da9c316dead254cb1c9c6caf9638bb23e96

salimbraksa: > This workaround seems critical for the correct attribution of events...

The addition of the `platform`  property only impacts **ExPlat**, Track events are still using the `eventNamePrefix` property.

Sources/Experiments/ExPlat.swift
```diff
@@ -30,6 +30,26 @@ import Cocoa
...
+        return service.oAuthToken
+    }
+
+    var userAgent: String? {
+        return service.userAgent
+    }
+
+    var anonId: String? {
+        return service.anonId
+    }
```
crazytonyli, line 49: I assume these properties are added for the new test cases to make assertions? Considering these new properties are not used any library code, IMO, it would be more robust to use `OHHTTPStubs` to intercept HTTP requests and assert that the requests contain appropriate values. What do you think?

mokagio: +1

Tests/Tests/ABTesting/ExPlatTests.swift
```diff
@@ -103,6 +116,58 @@ class ExPlatTests: XCTestCase {
...
+        // When
+        self.tracksService.switchToAnonymousUser(withAnonymousID: anonymousId)
+
+        // Then
+#if os(iOS)
+        let exPlat = ExPlat.shared
+        XCTAssertNotNil(exPlat)
+        XCTAssertEqual(exPlat?.platform, eventNamePrefix)
+        XCTAssertEqual(exPlat?.oAuthToken, nil)
+        XCTAssertEqual(exPlat?.anonId, anonymousId)
```
mokagio: I think this code will change following @crazytonyli's suggestion [above](https://github.com/Automattic/Automattic-Tracks-iOS/pull/253/files#r1179889461), but I just wanted to ask if you considered using `XCTUnwrap` to get what I think would be clearer test failures if `exPlat` was nil.

```suggestion
        let exPlat = try XCTUnwrap(ExPlat.shared)
        XCTAssertEqual(exPlat.platform, eventNamePrefix)
        XCTAssertEqual(exPlat.oAuthToken, nil)
        XCTAssertEqual(exPlat.anonId, anonymousId)
```

Tests/Tests/ABTesting/ExPlatTests.swift
```diff
@@ -2,18 +2,31 @@ import XCTest
...
+    var tracksService: TracksService!
+
+    override func setUp() {
+        let contextManager = MockTracksContextManager()
+        self.tracksService = TracksService(contextManager: contextManager)
+    }
+
+    override func tearDown() {
+        ExPlat.shared = nil
+    }
```
mokagio, line 28: Have you considered constraining the lifecycle of `trackService` to the test methods?

I consider this approach advantageous because:

1. When tests have a `setUp` method, it's easy to miss the information when looking at a single test
2. Because of the odd way XCTest works, that `tracksService` instance will be retained in memory _until the end of the entire test suite_, unless nilled in the `tearDown`.

I did something similar [in the Parsely iOS library](https://github.com/Parsely/AnalyticsSDK-iOS/pull/78/files#r1172093174) recently.

```swift
func makeTrackService() -> TracksService {
    let contextManager = MockTracksContextManager()
    let tracksService = TracksService(contextManager: contextManager)

    addTeardownBlock {
        ExPlat.shared = nil
    }
}

func testTracksServiceAnonymousConfiguration() {
    let trackService = makeTrackService()
    // ...
```

mokagio (commented):
Thank you for spending time to add tests @salimbraksa 🙇‍♂️ Just a few suggestions on top of what Tony's pointed out.

Is this new functionality something that will be needed in the upcoming 22.3 iOS version? If so, maybe we can followup with the tests refinements later on and merge as is (the test pass, after all 😄 ) to ship a new version

If that's the case, I can take care of that when doing the code freeze this coming Monday.

crazytonyli (approved):
As discussed internally, [this comment](https://github.com/Automattic/Automattic-Tracks-iOS/pull/253/files#r1179889461) will be addressed in follow PRs.

Tests/Tests/ABTesting/ExPlatTests.swift
salimbraksa: @mokagio Resolved in ec0af51dbd1c3b28876e995bfead11317f8815a2

PR #256, opened 2023-05-03
Refactor ExPlat assignments endpoint unit tests

salimbraksa:
This PR is to address the following feedback:

- [Use `OHHTTPStubs` to assert that the ExPlat assignments request contains appropriate values](https://github.com/Automattic/Automattic-Tracks-iOS/pull/253#discussion_r1179889461).
- [Constrain the lifecycle of `trackService` to the test methods](https://github.com/Automattic/Automattic-Tracks-iOS/pull/253#discussion_r1180275582).

---

### DIFF
--- a/Tests/Tests/ABTesting/ExPlatTests.swift
+++ b/Tests/Tests/ABTesting/ExPlatTests.swift
@@ -18,17 +18,6 @@ class ExPlatTests: XCTestCase {
         anonId: nil
     )
 
-    var tracksService: TracksService!
-
-    override func setUp() {
-        let contextManager = MockTracksContextManager()
-        self.tracksService = TracksService(contextManager: contextManager)
-    }
-
-    override func tearDown() {
-        ExPlat.shared = nil
-    }
-
     // Save the returned experiments variation
     //
     func testRefresh() {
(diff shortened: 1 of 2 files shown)


Tests/Tests/ABTesting/ExPlatTests.swift
```diff
@@ -117,54 +108,119 @@ class ExPlatTests: XCTestCase {
...
+        let path = "wpcom/v2/experiments/0.1.0/assignments"
+        var endpoint = "\(baseURL)/\(path)/\(platform)?_locale=en&experiment_names=\(experiment)"
+        if let anonId {
+            endpoint.append("&anon_id=\(anonId)")
+        }
+        return endpoint
+    }
+
+    private func makeTracksService(platform: String?, eventNamePrefix: String?) -> TracksService? {
+        guard let tracksService = TracksService(contextManager: MockTracksContextManager()) else {
```
crazytonyli, line 183: I don't think it's possible that the `tracksService` would be nil?

Tests/Tests/ABTesting/ExPlatTests.swift
```diff
@@ -117,54 +108,119 @@ class ExPlatTests: XCTestCase {
...
-        let eventNamePrefix = "jpios"
-        let skipAliasEventCreation = true
-        self.tracksService.platform = platform
-        self.tracksService.eventNamePrefix = eventNamePrefix
+        let expectation = XCTestExpectation(description: "The ExPlat assignments endpoint should have the right params")
+        let experiment = "test"
+        let platform = "wooios_test"
+        let token = "abc"
+        let expectedEndpoint = Self.makeAssignmentsEndpoint(platform: platform, experiment: experiment)
+        stub(condition: isAbsoluteURLString(expectedEndpoint)) { request in
```
crazytonyli, line 148: This condition is probably too strict. The query part in the `expectedEndpoint` probably won't exactly match the requested URL string. I think it's best to compare the path part in the condition, and assert the `request.url` contains certain query?

crazytonyli (changes requested):
Thanks for considering my suggestion. I think [these properties](https://github.com/Automattic/Automattic-Tracks-iOS/blob/be3d5c27840b7f575be58b520746bd8660ca7381/Sources/Experiments/ExPlat.swift#L35-L49) can now be removed?

Tests/Tests/ABTesting/ExPlatTests.swift
mokagio: See also [the matcher's implementation](https://github.com/AliSoftware/OHHTTPStubs/blob/12f19662426d0434d6c330c6974d53e2eb10ecd9/Sources/OHHTTPStubsSwift/OHHTTPStubsSwift.swift#L191).

PR #260, opened 2023-06-20
Expose an API to override Sentry's stitchAsyncCode option

salimbraksa:
Ref p1687271187638419-slack-C05CANZ2B6F

As the PR title says, we are adding the `options.stitchAsyncCode` during the Sentry initialization. It is `false` by default but clients can override that option.

This PR is needed for:
- https://github.com/wordpress-mobile/WordPress-iOS/pull/20912

At the time of writing this PR, there is a [newer option](https://github.com/getsentry/sentry-cocoa/pull/3051) in the latest Sentry version 8.8.0, named `swiftAsyncStacktraces`. However, this option didn't provide the expected results in the test cases outlined in https://github.com/wordpress-mobile/WordPress-iOS/pull/20912. In contrast, the `stitchAsyncCode` option delivered better results.

---

### DIFF
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -63,7 +63,7 @@ public class CrashLogging {
 
             /// Attach stack traces to non-fatal errors
             options.attachStacktrace = true
-            options.swiftAsyncStacktraces = self.dataProvider.swiftAsyncStacktraces
+            options.stitchAsyncCode = self.dataProvider.enhancedAsyncStacktraces
 
             // Performance monitoring options
             options.enableAutoPerformanceTracing = self.dataProvider.enableAutoPerformanceTracking
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -63,6 +63,7 @@ public class CrashLogging {
 
             /// Attach stack traces to non-fatal errors
             options.attachStacktrace = true
+            options.swiftAsyncStacktraces = self.dataProvider.swiftAsyncStacktraces
 
             // Performance monitoring options
             options.enableAutoPerformanceTracing = self.dataProvider.enableAutoPerformanceTracking
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -36,6 +36,6 @@ Pod::Spec.new do |s|
   s.module_name = 'AutomatticTracks'
 
   s.ios.dependency 'UIDeviceIdentifier', '~> 2.0'
-  s.dependency 'Sentry', '8.5.0'
+  s.dependency 'Sentry', '8.6.0'
   s.dependency 'Sodium', '>= 0.9.1'
 end
--- a/Automattic-Tracks-iOS.podspec
+++ b/Automattic-Tracks-iOS.podspec
@@ -36,6 +36,6 @@ Pod::Spec.new do |s|
   s.module_name = 'AutomatticTracks'
 
   s.ios.dependency 'UIDeviceIdentifier', '~> 2.0'
-  s.dependency 'Sentry', '~> 8.0'
+  s.dependency 'Sentry', '8.5.0'
   s.dependency 'Sodium', '>= 0.9.1'
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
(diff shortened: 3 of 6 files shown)


Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -63,6 +63,7 @@ public class CrashLogging {
 
             /// Attach stack traces to non-fatal errors
             options.attachStacktrace = true
+            options.stitchAsyncCode = self.dataProvider.enhancedAsyncStacktraces
```
mokagio, line 66: I like the "enhanced async stacktraces" as a name for this parameter.

mokagio (changes requested):
> At the time of writing this PR, there is a https://github.com/getsentry/sentry-cocoa/pull/3051 in the latest Sentry version 8.8.0, named swiftAsyncStacktraces. However, this option didn't provide the expected results in the test cases outlined in https://github.com/wordpress-mobile/WordPress-iOS/pull/20912. In contrast, the stitchAsyncCode option delivered better results.

I appreciated that implementation might have been working well during testing, however, [according to Sentry's own changelog](https://github.com/getsentry/sentry-cocoa/blob/257c2a998df2f9a0e257802d69d4268d4a83c560/CHANGELOG.md?plain=1#L102)

> [The `stitchAsyncCode`] behavior was unpredictable and sometimes resulted in unexpected errors.

As a matter of fact, the option was [removed from the library](https://github.com/getsentry/sentry-cocoa/pull/2973).

I'm afraid relying on this option will either result us facing the unexpected runtime behavior the Sentry team reported or put us in the difficult position of not being able to upgrade.

staskus:
Given `stitchAsyncCode` was removed due to instability, [should we then go with `swiftAsyncStacktraces`](https://github.com/getsentry/sentry-cocoa/pull/3051) instead? Or do you think it's still dangerous due to being experimental? They mention in PR:

> Switching from backtrace to backtrace_async is super trivial and gets Sentry 90% of the way to offering good Swift Concurrency support.

mokagio:
@staskus @salimbraksa 

> [should we then go with swiftAsyncStacktraces](https://github.com/getsentry/sentry-cocoa/pull/3051) instead? Or do you think it's still dangerous due to being experimental?

The feature being experimental is a liability, in that they might roll it back at some point forcing us to update our code. But, given that's their latest version it seems a more secure approach than using a setting they _already_ removed.

staskus:
> The feature being experimental is a liability, in that they might roll it back at some point forcing us to update our code

True, it is still an experimental and early feature.

> Given that's their latest version it seems a more secure approach than using a setting they already removed.

As far as I understood, this option seems to be safe. For non-swift-async use cases, internals behaves the same way as before:

> [backtrace_async](https://keith.github.io/xcode-man-pages/backtrace.3.html#backtrace_async)() behaves exactly like backtrace() unless it is invoked from a Swift async context. In that case, instead of writing the return addresses of the OS call stack, the continuation addresses of the async invocations that led to the current state are provided.
-- https://keith.github.io/xcode-man-pages/backtrace.3.html#backtrace_async

However, as, @salimbraksa mentioned, this option is not as effective now:

> At the time of writing this PR, there is a https://github.com/getsentry/sentry-cocoa/pull/3051 in the latest Sentry version 8.8.0, named swiftAsyncStacktraces. However, this option didn't provide the expected results in the test cases outlined in https://github.com/wordpress-mobile/WordPress-iOS/pull/20912. In contrast, the stitchAsyncCode option delivered better results.

AFAIK, it's only designed to work for native Swift concurrency (async/await). There aren't many places in the code using it yet. 

Given the feature is experimental and introduced in the latest Sentry version, and is not likely to provide many insights at the moment, I say we wait for a newer Sentry version when this [Swift async support is more solid](https://github.com/getsentry/sentry-cocoa/issues/304).

@mokagio, @salimbraksa

salimbraksa:
> Given the feature is experimental and introduced in the latest Sentry version, and is not likely to provide many insights at the moment, I say we wait for a newer Sentry version when this https://github.com/getsentry/sentry-cocoa/issues/304.

That sounds good to me. My initial intention was to use the experimental `stitchAsyncCode` as a last resort if none of the other solutions worked. And I intended to use it like this:

1. Enable this feature only in the Beta app, and invite some users ( or Matt ) to use the Beta build.
2. Enable this feature exclusively for users who are consistently encountering crashes. By doing so, users who have never experienced crashes will remain unaffected by the potential instability of this experimental feature.

Given the reduced occurrence of `__NSCFSet` crashes following the 22.5.1.0 release ( see p1688004316907819-slack-C05CANZ2B6F ), this positive trend indicates that we may not currently need these experimental features. Therefore, I will close this PR for now, and we can reconsider if necessary at a later time.

@staskus @mokagio

PR #261, opened 2023-08-02
Add options to config Sentry app hang and HTTP client error tracking

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/10372

## Description
Since Sentry v8.0.0, app hang and HTTP client errors are tracked by default. These reports overflow the issue list for WCiOS, so we want to add an option to disable these tracking from our side.

This PR adds new options for `CrashLoggingDataProvider`: `enableAppHangTracking` and `enableCaptureFailedRequests`. These are used to configure relevant Sentry options. They are kept as ~`true`~ `false` by default ~not to affect any existing apps using this dependency~ and other apps can opt in if needed.

---

### DIFF
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -13,6 +13,10 @@ public protocol CrashLoggingDataProvider {
     var additionalUserData: [String: Any] { get }
     var shouldEnableAutomaticSessionTracking: Bool { get }
     var performanceTracking: PerformanceTracking { get }
+    /// Whether app hang are captured.
+    var enableAppHangTracking: Bool { get }
+    /// Whether HTTP client errors are captured.
+    var enableCaptureFailedRequests: Bool { get }
 }
 
 /// Default implementations of common protocol properties
@@ -79,5 +83,15 @@ public extension CrashLoggingDataProvider {
     var enableUserInteractionTracing: Bool {
         guard case .enabled(let config) = performanceTracking else { return false }
         return config.trackUserInteraction
+    }
+
+    /// App hang tracking is enabled by default.
+    var enableAppHangTracking: Bool {
+        return true
+    }
+
+    /// HTTP client errors are captured by default.
+    var enableCaptureFailedRequests: Bool {
+        return true
     }
 }
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -85,13 +85,13 @@ public extension CrashLoggingDataProvider {
         return config.trackUserInteraction
     }
 
-    /// App hang tracking is enabled by default.
+    /// App hang tracking is disabled by default to avoid unexpected events being tracked.
     var enableAppHangTracking: Bool {
-        return true
+        return false
     }
 
-    /// HTTP client errors are captured by default.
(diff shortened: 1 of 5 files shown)


itsmeichigo:
@mokagio I'm setting the new version with `2.3.0` as we are not planning to ship a new version soon for this library. I'm still confused about the versioning since the [last update](https://github.com/wordpress-mobile/WordPressUI-iOS/pull/128) of `WordPressUI`, so pinging you to make sure that this is ok.

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -80,4 +84,14 @@ public extension CrashLoggingDataProvider {
...
+
+    /// App hang tracking is enabled by default.
+    var enableAppHangTracking: Bool {
+        return true
+    }
+
+    /// HTTP client errors are captured by default.
+    var enableCaptureFailedRequests: Bool {
+        return true
+    }
```
mokagio: I appreciated the idea of maintaining the default Sentry behavior to avoid breaking the experience of existing apps. At the same time, making it opt-in would ensure that other apps won't experience the same clutter you noted. What do you think?

```suggestion
    /// App hang tracking is disabled by default to avoid unexpected events being tracked.
    var enableAppHangTracking: Bool {
        return false
    }

    /// HTTP client errors are disabled by default to avoid unexpected events being tracked.
    var enableCaptureFailedRequests: Bool {
        return false
    }
```

mokagio (approved):

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
mokagio: If you accept this change, I'll publish a PSA post after releasing the new version advising folks to opt-in if they need either log.

itsmeichigo: I think that would be great! Accepting the suggested change now.

itsmeichigo:
@mokagio I merged the change, could you help with the release, please? 🙇 
Edited: I created a release for 2.3.0 myself but couldn't trigger CI to push a new version of the library on Cocoapods. I'll proceed to merge the WCiOS PR pointing to `trunk` of this library for now.

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
kean: The WP/JP team uses app hang tracking and finds it useful. But it's often inaccurate. As @salimbraksa suggested (p1691161246764459-slack-C05CANZ2B6F), I think we should look for other options. Apple's [MetricKit](https://developer.apple.com/documentation/metrickit) has hang diagnostics and more.

kean: From [SentryOptions.m](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentryOptions.m)

```swift
self.enableAppHangTracking = YES;
self.enableCaptureFailedRequests = YES;
```

Both app hangs and capture failed requests are enabled by Sentry by default. With this change, they are now going to be disabled.

kean: `enableCaptureFailedRequests` – it sounds like a good idea to disable these for good. It should typically be the server monitoring its health, not the clients. It should reduce the amount of swizzling and some overhead on the client.

mokagio:
@itsmeichigo 👍 👍 👍

PR #267, opened 2023-11-11
Update `releaseName` default value

spencertransier:
This PR updates the default `releaseName` attribute from sending only the `CFBundleVersionKey` to now include the `CFBundleIdentifierKey`, `CFBundleShortVersionString`, and `CFBundleVersionKey`. Sentry releases are global across all projects in an organization, so only relying on the `CFBundleVersionKey` was causing conflicts between apps that had the same release number. This will now correctly separate the releases.

The `releaseName` format that we're sending now is the default format that Sentry uses for the `releaseName`. I decided that it would be good to explicitly define that format in this library so that we can ensure future consistency for release names across all the apps. 

I had trouble using the TracksDemo app to test updating the release version in Sentry, so I created a test branch for WooCommerce iOS that points to this dev branch: `test/sentry-release-name` (The only change there is updating the `podfile` to point here)

Using that test branch, I was able to verify that the releases are now being sent up to Sentry with the updated attributes. You can see that the `16.2 (16.2.0.0)` and `16.1 (16.1.0.1)` releases show the full version information now:

Screenshot 2023-11-13 at 4 43 21 PM

---

### DIFF
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -22,11 +22,6 @@ public protocol CrashLoggingDataProvider {
 /// Default implementations of common protocol properties
 public extension CrashLoggingDataProvider {
 
-    // According to https://docs.sentry.io/platforms/apple/configuration/releases/:
-    // "If no release name is set, the SDK creates a default combined from CFBundleIdentifier,
-    // CFBundleShortVersionString, and CFBundleVersion, for example my.project.name@2.3.12+1234"
-    //
-    // Clients can set a custom releaseName value if needed
     var releaseName: String {
         let bundleVersion = Bundle.main.infoDictionary?[kCFBundleVersionKey as String] ?? ""
         let bundleIdentifier = Bundle.main.infoDictionary?[kCFBundleIdentifierKey as String] ?? ""
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -23,11 +23,11 @@ public protocol CrashLoggingDataProvider {
 public extension CrashLoggingDataProvider {
 
     // According to https://docs.sentry.io/platforms/apple/configuration/releases/:
-    // "If no release name is set, the SDK creates a default combined from CFBundleIdentifier, 
+    // "If no release name is set, the SDK creates a default combined from CFBundleIdentifier,
     // CFBundleShortVersionString, and CFBundleVersion, for example my.project.name@2.3.12+1234"
     //
     // Clients can set a custom releaseName value if needed
-    var releaseName: String? = nil
+    var releaseName: String?
 
     var additionalUserData: [String: Any] {
         return [ : ]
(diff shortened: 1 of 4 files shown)


Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -23,7 +23,13 @@ public protocol CrashLoggingDataProvider {
...
     var releaseName: String {
-        return Bundle.main.object(forInfoDictionaryKey: kCFBundleVersionKey as String) as! String
+        let bundleVersion = Bundle.main.infoDictionary?[kCFBundleVersionKey as String] ?? ""
+        let bundleIdentifier = Bundle.main.infoDictionary?[kCFBundleIdentifierKey as String] ?? ""
+        let bundleShortVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? ""
+
+        // This is the format that Sentry recommends for version numbers.
+        // See https://docs.sentry.io/platforms/apple/configuration/releases/#bind-the-version
+        return "\(bundleIdentifier)@\(bundleShortVersion)+\(bundleVersion)"
     }
```
mokagio: Looking at the docs from that link, Sentry says it can set the value by itself if none is provided:

image

Have you considered doing that?

mokagio (commented):
How do you feel about adding a unit test to verify the format we compute is as we expect it to be right now and ensure it will remain like that in the future (or, rather, have a safeguard against unintentionally changing it)

mokagio:
@spencertransier I saw CI failed because the change in the source code now requires changes to the tests too, which makes me wonder whether it might require changes in the clients as well.

Depending on the amount of work required, and given that the release tracking inconsistency is a blocker for a project in WooCommerce (internal ref pdnsEh-1mT-p2) feel free to ship a version with your override since you already proven it works. We can iterate on this with more clam later. 👍

spencertransier:
@mokagio When I was chatting with @jkmassel about this the other day, he suggested making `releaseName` an optional string so that clients could pass in their own release name if needed. Looking through our iOS apps, though, DOiOS/Mac is the only one not relying on the default behavior that's current the Tracks library. And even DO is [just passing the same format as the Sentry default](https://github.com/bloom/DayOne-Apple/blob/638e7285c86e03f86c00b39f275dda52726e1bc4/core/DOCore/DOCore/DOCrashLogging.swift#L74C1-L74C1) so that the DO releases would be configured correctly (what WCiOS is trying to accomplish). Seeing that all the apps want this format anyway, I don't see a need to provide a way to customize the `releaseName` in this library.

Considering that, what do you think about the changes I just pushed to remove the `releaseName` from the protocol entirely and only rely on Sentry's default? It would be a breaking change and we would need to bump the library to `3.0.0`. DOiOS/Mac is the only A8C app that would actually need to be updated (and it would just be removing their custom releaseName at that). Then the other A8C apps that update to `3.0.0` would get the improved `releaseName`'s for free without needing to making any changes other than updating the library.

I tried out the latest PR change in the WCiOS test branch and it worked great. The new version showed up in Sentry with the expected format.

mokagio (approved):
> Considering that, what do you think about the changes I just pushed to remove the `releaseName` from the protocol entirely and only rely on Sentry's default?

I like the idea!

> I tried out the latest PR change in the WCiOS test branch and it worked great. The new version showed up in Sentry with the expected format.

Awesome.

> It would be a breaking change and we would need to bump the library to 3.0.0. 

Would it be a breaking change? The `var { get }` requirement has been removed from the protocol, which makes me think that clients implementing it would not break as the compiler would consider their property declaration to be one for the object itself and not required by the protocol. If we had added one, then that's when I think the compiler would bark at us.

Regardless of the version number, let's do this!

PR #271, opened 2023-12-07
Exposing sampleRate parameter

jleandroperez:
### Details:
In this PR we're enhancing `CrashLoggingDataProvider`: our goal is for the host app to specify the `sampleRate` in which events are uploaded.

We're gonna be using the Events subsystem to track down sync errors (starting at a 1% sample rate).

Thanks in advance!!

---

### DIFF
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -66,7 +66,7 @@ public class CrashLogging {
             options.attachStacktrace = true
 
             // Events
-            options.sampleRate = NSNumber(value: self.dataProvider.errorEventsSamplingRate)
+            options.sampleRate = NSNumber(value: min(max(self.dataProvider.errorEventsSamplingRate, 0), 1))
 
             // Performance monitoring options
             options.enableAutoPerformanceTracing = self.dataProvider.enableAutoPerformanceTracking
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -66,7 +66,7 @@ public class CrashLogging {
             options.attachStacktrace = true
 
             // Events
-            options.sampleRate = NSNumber(value: self.dataProvider.eventSamplingRate)
+            options.sampleRate = NSNumber(value: self.dataProvider.errorEventsSamplingRate)
 
             // Performance monitoring options
             options.enableAutoPerformanceTracing = self.dataProvider.enableAutoPerformanceTracking
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -65,6 +65,9 @@ public class CrashLogging {
             /// Attach stack traces to non-fatal errors
             options.attachStacktrace = true
 
+            // Events
+            options.sampleRate = NSNumber(value: self.dataProvider.eventSamplingRate)
+
             // Performance monitoring options
             options.enableAutoPerformanceTracing = self.dataProvider.enableAutoPerformanceTracking
             options.tracesSampler = { _ in
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -92,6 +92,6 @@ public extension CrashLoggingDataProvider {
     }
 
     var errorEventsSamplingRate: Double {
-        return 0.01
+        return 1.0
     }
 }
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -10,7 +10,7 @@ public protocol CrashLoggingDataProvider {
     var buildType: String { get }
     var currentUser: TracksUser? { get }
     var additionalUserData: [String: Any] { get }
-    var eventSamplingRate: Double { get }
+    var errorEventsSamplingRate: Double { get }
     var shouldEnableAutomaticSessionTracking: Bool { get }
     var performanceTracking: PerformanceTracking { get }
     /// Whether app hang are captured.
@@ -91,7 +91,7 @@ public extension CrashLoggingDataProvider {
         return false
     }
 
-    var eventSamplingRate: Double {
+    var errorEventsSamplingRate: Double {
         return 0.01
     }
 }
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -10,6 +10,7 @@ public protocol CrashLoggingDataProvider {
     var buildType: String { get }
     var currentUser: TracksUser? { get }
     var additionalUserData: [String: Any] { get }
+    var eventSamplingRate: Double { get }
     var shouldEnableAutomaticSessionTracking: Bool { get }
     var performanceTracking: PerformanceTracking { get }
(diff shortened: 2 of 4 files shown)


jleandroperez:
Gentleman, sorry to bug you all!! Adding an extra parameter, so that we can control the rate at which Events are reported.

Context: We'd like to use this to shed some light into DayOne's sync errors. https://github.com/bloom/DayOne-Apple/issues/22360

Thank you all in advance!!

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -10,6 +10,7 @@ public protocol CrashLoggingDataProvider {
     var buildType: String { get }
     var currentUser: TracksUser? { get }
     var additionalUserData: [String: Any] { get }
+    var eventSamplingRate: Double { get }
```
mokagio: In the diff above, this property maps to Sentry's `sampleRate`. [The docs](https://docs.sentry.io/platforms/apple/guides/ios/configuration/sampling/#sampling-error-events) refer to that property in the context of "sampling error events".

What do you think of renaming it `errorEventsSamplingRate`?

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -89,4 +90,8 @@ public extension CrashLoggingDataProvider {
     var enableCaptureFailedRequests: Bool {
         return false
     }
+
+    var eventSamplingRate: Double {
+        return 0.01
```
mokagio: From [the Sentry docs](https://docs.sentry.io/platforms/apple/guides/ios/configuration/sampling/#sampling-error-events): "The error sample rate defaults to 1, meaning all errors are sent to Sentry."

Setting this as `0.01` by default might result in clients that are not aware of this new configuration to lose 99% of their error events.

What do you think of keeping the same default as Sentry?

```suggestion
        return 1.0
```

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
mokagio: I assume Sentry would clamp this value between 0 and 1 internally, but what do you think of doing the same on our end, similar to what we do fro performance tracking:

https://github.com/Automattic/Automattic-Tracks-iOS/blob/d883cd1720ca49c2f2ec7e8473c1f49eb0df0c4b/Tests/Tests/PerformanceTrackingTests.swift#L11-L31

mokagio (commented):
I have nothing against making the error events sample rate configurable at the library's end. But, just in case, have you considered using Sentry's backend [rate limiting](https://docs.sentry.io/product/accounts/quotas/manage-attachments-quota/#rate-limiting) to mitigate lots of errors being logged? Or is that not granular enough for your needs?

From [the docs](https://docs.sentry.io/platforms/apple/guides/ios/configuration/sampling/#sampling-error-events):

> Changing the error sample rate requires re-deployment. In addition, setting an SDK sample rate limits visibility into the source of events. Setting a rate limit for your project (which only drops events when volume is high) may better suit your needs.

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
jleandroperez: Sounds super, thank you Gio!!

Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
jleandroperez: Done, thank you!!!

jleandroperez:
Hola @mokagio!!! hope you're doing super!!

All feedback addressed. Replicating same note as in https://github.com/bloom/DayOne-Apple/issues/22360, for visibility.

All I could find in the **[events filter (web settings) interface](https://a8c.sentry.io/settings/projects/day-one-ios/filters/data-filters/)** are filters related to User Agent, Releases or specific IP(s).

Unless I'm missing something huuuuge, perhaps it's not possible to replicate the same thing this setting does? (Events Sampling Rate).

Thanks in advance!!

mokagio:
> perhaps it's not possible to replicate the same thing this setting does? (Events Sampling Rate).

I noticed there's a way to [filter events based on the error message](https://docs.sentry.io/product/data-management-settings/filtering/#error-message):

image

But it looks like an on-off switch, which might be good as a killswitch in case of a spike, but not good to have a fraction of the events reported for monitoring.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -65,6 +65,9 @@ public class CrashLogging {
             /// Attach stack traces to non-fatal errors
             options.attachStacktrace = true
 
+            // Events
+            options.sampleRate = NSNumber(value: min(max(self.dataProvider.errorEventsSamplingRate, 0), 1))
```
mokagio, line 69: Right. I suggested clamping this and was expecting to see tests for the behavior, but forgot that `CrashLoggingDataProvider` is a protocol. So it's really up to the type in the client that will conform to it to implement the clamping if they see fit. Still, it's great to have the safety check here, too. Thanks!

I suppose we could extract the `CrashLogging` logic that goes from `CrashLoggingDataProvider` to `SentryOptions`, and test that, but it seems out of scope.

Tests/Tests/CrashLoggingTests.swift
```diff
@@ -130,6 +130,7 @@ class CrashLoggingTests: XCTestCase {
 
     func testPerformanceMonitoringIsDisabledByDefault() {
         let dataProvider = CrashLoggingDataProviderWithDefaultValuesOnly(
+            errorEventsSamplingRate: 0.01,
```
mokagio: The idea of this test is to only pass require parameters then to make assertions on how the ones that have default values specified via protocol extension behave.

That is, omit this setting and then add `XCTAssertEqual(dataProvider.errorEventsSamplingRate, 1.0)`

Tests/Tests/CrashLoggingTests.swift
```diff
@@ -223,6 +225,7 @@ private class MockCrashLoggingDataProvider: CrashLoggingDataProvider {
 
 /// Use this to get a `CrashLoggingDataProvider` implementation that allows testing the default values set via protocol extenstion.
 private struct CrashLoggingDataProviderWithDefaultValuesOnly: CrashLoggingDataProvider {
+    var errorEventsSamplingRate: Double
```
mokagio: Following up on the comment above regarding testing that the default value for `errorEventsSamplingRate` is `1.0`, this configurable property is not necessary.

```suggestion
/// Use this to get a `CrashLoggingDataProvider` implementation that allows testing the default values set via protocol extenstion.
private struct CrashLoggingDataProviderWithDefaultValuesOnly: CrashLoggingDataProvider {
```

mokagio (approved):
Looks like there's really no backend way out of this, which is unfortunate because dialing down the error events sample rate will require in undiscriminated data loss. But hey, making this configurable means we can turn it back to 1.0 when satisfied.

Only one minor comment about the tests. Approving to unblock in the mean time.

Tests/Tests/CrashLoggingTests.swift
jleandroperez: Thank you so much for picking up on this, I absolutely missed the point of the test (just skimmed thru and added the new default!)

Tests/Tests/CrashLoggingTests.swift
jleandroperez: Thank you Gio!!

jleandroperez:
@mokagio I swear I started by writing tests, but I immediately noticed that there is no direct way to query those values, and yeah I'm afraid we're a bit tight on time, and extracting the SentryOptions initializer is a bit out of scope atm.

**Thanks a lot for your feedback!!!**

PR #272, opened 2023-12-14
Bump version to 3.1.0

mokagio:
This version bump PR is here only to leave a breadcrumb in the release process. Once CI is green, I will push the tag for this version and, once that is successfully deployed, admin-merge this.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 # Nodes with values to reuse in the pipeline.
 common_params:
   plugins: &common_plugins
-    - automattic/a8c-ci-toolkit#2.17.0
+    - automattic/a8c-ci-toolkit#3.0.1
   # Common environment values to use with the `env` key.
   env: &common_env
     IMAGE_ID: xcode-15.0.1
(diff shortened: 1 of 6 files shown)


mokagio:
CI failed to validate the pod because of deprecation warnings:

```
- WARN  \| xcodebuild:  /opt/ci/builds/builder/automattic/tracks-ios/Sources/Model/ObjC/Common/TracksDeviceInformation.m:177:48: warning: 'statusBarFrame' is deprecated: first deprecated in iOS 13.0 - Use the statusBarManager property of the window scene instead. [-Wdeprecated-declarations]
```

I'll deploy the release manually from my machine using the `--allow-warnings` flag.

I should be straightforward to update the code to use the new recommended syntax ([see StackOverflow, for example](https://stackoverflow.com/questions/57023899/how-to-get-the-status-bar-height-in-ios-13)) but I don't want to do it in the context of a release PR that I will admin merge.

mokagio:
> I'll deploy the release manually from my machine using the --allow-warnings flag.

Taking longer than I wished because of slow Australian internet...

Screenshot 2023-12-14 at 14 52 42

Issue #273, opened 2023-12-14
Remove deprecated `statusBarFrame` usages


mokagio:
The 3.1.0 CI deployment failed because of deprecation warnings:

```
- WARN  \| xcodebuild:  /opt/ci/builds/builder/automattic/tracks-ios/Sources/Model/ObjC/Common/TracksDeviceInformation.m:177:48:
  warning: 'statusBarFrame' is deprecated: first deprecated in iOS 13.0
    - Use the statusBarManager property of the window scene instead. [-Wdeprecated-declarations]
```

We could work around this by passing `--allow-warnings` to the pod validation command via our CI plugin, but I think it's best we remain strict in our validation and address this deprecation. It's deprecated for a reason, after all.

It should be straightforward to update the code to use the new recommended syntax ([see StackOverflow, for example](https://stackoverflow.com/questions/57023899/how-to-get-the-status-bar-height-in-ios-13)).

crazytonyli:
This `statusBarFrame` is used to [track status bar height](https://github.com/Automattic/Automattic-Tracks-iOS/blob/3.2.0/Sources/Event%20Logging/TracksService.m#L413). But is that information really useful to us? If it's not, can we just delete it?

There is also another similar deprecated API usage: [`statusBarOrientation`](https://github.com/Automattic/Automattic-Tracks-iOS/blob/b6979ef69b4b094c8809ba83661222dcd0d7667e/Sources/Model/ObjC/Common/TracksDeviceInformation.m#L196). It's used to track device orientation. I think this information can be useful. But I think it's more appropriate to use [`UIDevice.orientation`](https://developer.apple.com/documentation/uikit/uidevice/1620053-orientation) instead.

bjhomer:
Another problem: both of these APIs can only accessed on the main thread. This means that every Tracks call from a non-main thread is making a _blocking_ call to the main thread to fetch the status bar frame.

If nobody's using this information, I'd agree with removing it.

PR #274, opened 2023-12-16
CrashLogging: New Error Logging API

jleandroperez:
### Details:

We'll be using Sentry to [track down Sync-Errors in DayOne](https://github.com/bloom/DayOne-Apple/pull/22377). While testing the main integration PR, I've noticed **it was next to impossible** to filter by a given event's **Additional Data** fields. 

You can find a [Sample Event Here](https://a8c.sentry.io/issues/4729708951/?project=5959376&query=user.id%3A1677223025&referrer=issue-stream&statsPeriod=7d&stream_index=2), with the `Additional Data` set (but no Tags)

Since Sentry recommends using Enriched Events, which can then be [filtered with the keyword `has:`.](https://a8c.sentry.io/issues/?project=5959376&query=has%3Async-error&referrer=issue-list&statsPeriod=14d), in this PR we're adding an extra API that allows us to set a `Tag/Value` pair in the event's context.

cc @jkmassel  (Thanks a lot in advance!!

---

### DIFF
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -146,7 +146,28 @@ public class CrashLogging {
 // MARK: - Manual Error Logging
 public extension CrashLogging {
 
+    /// Writes the error to the Crash Logging system, and includes a stack trace. This API supports for rich events.
+    /// By setting a Tag/Value pair, you'll be able to filter these events, directly, with the `has:` operator (Sentry Web Interface).
     ///
+    /// - Parameters:
+    ///   - error: The error object
+    ///   - tagName: Tag name to be associated with the error
+    ///   - tagValue: Tag's value to be associated with the error
+    ///   - level: The level of severity to report in Sentry (`.error` by default)
+    func logError(_ error: Error, tagName: String, tagValue: String, level: SentryLevel = .error) {
+
+        let event = Event.from(
+            error: error as NSError,
+            level: level
+        )
+
+        SentrySDK.capture(event: event) { scope in
+            scope.setTag(value: tagValue, key: tagName)
+        }
+
+        dataProvider.didLogErrorCallback?(event)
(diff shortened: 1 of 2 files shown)


Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -146,7 +146,28 @@ public class CrashLogging {
...
+            level: level
+        )
+
+        SentrySDK.capture(event: event) { scope in
+            scope.setTag(value: tagValue, key: tagName)
+        }
+
+        dataProvider.didLogErrorCallback?(event)
+    }
+
```
jleandroperez, line 171: No easy way to unit test this. I've tinkered with the idea of adding `didLogTaggedErrorCallback`, but TBH the entire test felt pointless, since it would only be able to evaluate that the callback gets invoked 😆

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -146,7 +146,30 @@ public class CrashLogging {
 // MARK: - Manual Error Logging
 public extension CrashLogging {
 
+    /// Writes the error to the Crash Logging system, and includes a stack trace. This API supports for rich events.
+    /// By setting a Tag/Value pair, you'll be able to filter these events, directly, with the `has:` operator (Sentry Web Interface).
     ///
+    /// - Parameters:
+    ///   - error: The error object
+    ///   - tagName: Tag name to be associated with the error
+    ///   - tagValue: Tag's value to be associated with the error
```
crazytonyli: The variables here (`tagName` & `tagValue`) don't match the function.

crazytonyli (approved):

jleandroperez:
CI seems to be failing because of a couple new warnings, unrelated to this PR.
Those are being tracked in #273.

Merging this one back into Trunk. Thanks a lot Tony!!

PR #278, opened 2024-02-20
Add support to log JavaScript exceptions to Sentry

fluiddot:
**Related PRs:**
* https://github.com/WordPress/gutenberg/pull/59221
* https://github.com/wordpress-mobile/gutenberg-mobile/pull/6655
* https://github.com/wordpress-mobile/WordPress-iOS/pull/22655

This PR expands the Crash logging service to support logging JavaScript exceptions to Sentry. This will allow hybrid sources like React Native (used in Gutenberg Mobile) to log exceptions with detailed information and symbolicated stack traces, which will greatly simplify the crash debugging process.

The approach is based on previous attempts to log JavaScript exceptions (p9ugOq-249-p2). However, in this case, we are following an SDK-less implementation where the exception is bubbled up to the host app and sent as a Sentry event using the Sentry iOS SDK. As can be seen in the implementation, the Sentry event is adjusted to ensure that Sentry processes it as a JavaScript exception and hence, symbolicate the stack trace with previously uploaded source maps.

The process to log a JavaScript exception has the following steps:
- Gutenberg:
    - [Captures exceptions via error boundary components and a global error handler.](https://github.com/wordpress-mobile/gutenberg-mobile/blob/add/error-boundary/src/errorLogging/index.js)
    - [Parses the exception to convert the format of the stack trace from plain text to an array.](https://github.com/WordPress/gutenberg/blob/rnmobile/add/error-boundary/packages/react-native-bridge/index.js#L537-L591) ([parser](https://github.com/WordPress/gutenberg/blob/rnmobile/add/error-boundary/packages/react-native-bridge/lib/parseException.js))
    - [Sends the parsed exception to the host app (e.g. Jetpack app).](https://github.com/WordPress/gutenberg/blob/rnmobile/add/error-boundary/packages/react-native-bridge/ios/RNReactNativeGutenbergBridge.swift#L430-L441)
 - Host app:
    -  [Logs the JavaScript exception to the Crash logging service.](https://github.com/wordpress-mobile/WordPress-iOS/blob/rnmobile/add/log-exception-to-crash-logging/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L1357-L1376)
 - Crash logging service:
    -  [Creates a Sentry event with the JavaScript exception information, ensuring that it will be reported as a JavaScript exception in Sentry.](https://github.com/Automattic/Automattic-Tracks-iOS/blob/add/sentry-js-exception/Sources/Remote%20Logging/Crash%20Logging/SentryEventJSException.swift)

Regarding the source maps, they are uploaded as part of the build process of the host app ([reference](https://github.com/wordpress-mobile/WordPress-iOS/blob/rnmobile/add/log-exception-to-crash-logging/fastlane/lanes/build.rb#L485-L520)).

_More information about this approach is detailed in p9ugOq-4p6-p2._

## To test
Since it involves testing exceptions, we need to modify the code to force them and generate an installable build. A test PR has been created for this purpose, follow the testing instructions from https://github.com/wordpress-mobile/gutenberg-mobile/pull/6654.

---

### DIFF
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -152,25 +152,9 @@ public extension CrashLogging {
     /// - Parameters:
     ///   - exception: The exception object
     ///   - callback: Callback triggered upon completion
-    func logJavaScriptException(_ exception: [AnyHashable: Any], callback: @escaping () -> Void) {
-        let jsException = SentryEventJSException.initWithException(exception)
+    func logJavaScriptException(_ jsException: JSException, callback: @escaping () -> Void) {
         
-        if jsException.context == nil {
-            jsException.context = [:]
-        }
-        
-        var reactNativeContext:[String: Any] = exception["context"] as! [String: Any] ?? [:]
-        jsException.context?["react_native_context"] = reactNativeContext;
-        
-        if jsException.tags == nil {
-            jsException.tags = [:]
-        }
-        
-        if exception["tags"] != nil {
-            jsException.tags = jsException.tags?.merging(exception["tags"] as! [String: String]) { $1 }
-        }
-        
-        SentrySDK.capture(event: jsException)
+        SentrySDK.capture(event: SentryEventJSException.initWithException(jsException))
         
         DispatchQueue.global().async {
             SentrySDK.flush(timeout: self.flushTimeout)
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -146,7 +146,13 @@ public class CrashLogging {
 // MARK: - Manual Error Logging
 public extension CrashLogging {
 
-    func logJavaScriptException(_ exception: [AnyHashable: Any]) {
+    /// Writes a JavaScript exception to the Crash Logging system, including its stack trace.
+    /// Note that this function is provided mainly for hybrid sources like React Native.
+    ///
+    /// - Parameters:
+    ///   - exception: The exception object
+    ///   - callback: Callback triggered upon completion
+    func logJavaScriptException(_ exception: [AnyHashable: Any], callback: @escaping () -> Void) {
         let jsException = SentryEventJSException.initWithException(exception)
         
         if jsException.context == nil {
@@ -165,6 +171,11 @@ public extension CrashLogging {
         }
         
         SentrySDK.capture(event: jsException)
+        
+        DispatchQueue.global().async {
+            SentrySDK.flush(timeout: self.flushTimeout)
+            callback()
+        }
     }
     
     /// Writes the error to the Crash Logging system, and includes a stack trace. This API supports for rich events.
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -154,7 +154,7 @@ public extension CrashLogging {
     ///   - callback: Callback triggered upon completion
     func logJavaScriptException(_ jsException: JSException, callback: @escaping () -> Void) {
         
-        SentrySDK.capture(event: SentryEventJSException.initWithException(jsException))
+        SentrySDK.capture(event: SentryEventJSException.init(jsException: jsException))
         
         DispatchQueue.global().async {
             SentrySDK.flush(timeout: self.flushTimeout)
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -152,7 +152,7 @@ public extension CrashLogging {
     /// - Parameters:
     ///   - exception: The exception object
     ///   - callback: Callback triggered upon completion
-    func logJavaScriptException(_ jsException: JSException, callback: @escaping () -> Void) {
+    func logJavaScriptException(_ jsException: any JSException, callback: @escaping () -> Void) {
         
         SentrySDK.capture(event: SentryEventJSException.init(jsException: jsException))
         
--- a/Sources/Remote Logging/Crash Logging/CrashLogging.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLogging.swift
@@ -146,6 +146,27 @@ public class CrashLogging {
 // MARK: - Manual Error Logging
 public extension CrashLogging {
 
+    func logJavaScriptException(_ exception: [AnyHashable: Any]) {
+        let jsException = SentryEventJSException.initWithException(exception)
+        
+        if jsException.context == nil {
+            jsException.context = [:]
+        }
+        
+        var reactNativeContext:[String: Any] = exception["context"] as! [String: Any] ?? [:]
+        jsException.context?["react_native_context"] = reactNativeContext;
+        
+        if jsException.tags == nil {
+            jsException.tags = [:]
+        }
+        
+        if exception["tags"] != nil {
+            jsException.tags = jsException.tags?.merging(exception["tags"] as! [String: String]) { $1 }
+        }
+        
+        SentrySDK.capture(event: jsException)
+    }
+    
     /// Writes the error to the Crash Logging system, and includes a stack trace. This API supports for rich events.
     /// By setting a Tag/Value pair, you'll be able to filter these events, directly, with the `has:` operator (Sentry Web Interface).
     ///
--- a/Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
+++ b/Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
@@ -44,8 +44,8 @@ public class SentryEventJSException: Event {
         self.platform = "javascript"
     }
     
-    public static func initWithException(_ jsException: JSException) -> SentryEventJSException {
-        let sentryEvent = self.init()
+    public convenience init(jsException: JSException) {
+        self.init()
         
         // Generate exception based on JavaScript exception parameters
         let sentryException = Exception(value: jsException.value, type: jsException.type)
@@ -68,18 +68,16 @@ public class SentryEventJSException: Event {
         sentryException.mechanism = mechanism
         
         // Attach JavaScript exception to Sentry event
-        sentryEvent.exceptions = [sentryException]
+        self.exceptions = [sentryException]
         
         // Set event context
-        var context = sentryEvent.context ?? [:]
+        var context = self.context ?? [:]
         context["react_native_context"] = jsException.context;
-        sentryEvent.context = context
+        self.context = context
         
         // Set event tags
-        let tags = sentryEvent.tags ?? [:]
-        sentryEvent.tags = tags.merging(jsException.tags) { $1 }
-        
-        return sentryEvent
+        let tags = self.tags ?? [:]
+        self.tags = tags.merging(jsException.tags) { $1 }
     }
     
     override public func serialize() -> [String : Any] {
--- a/Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
+++ b/Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
@@ -3,16 +3,16 @@ import Sentry
 
 public struct JSException {
     public let type: String
-    public let value: String
+    public let message: String
     public let stacktrace: [StacktraceLine]
     public let context: [String: Any]
     public let tags: [String: String]
     public let isHandled: Bool
     public let handledBy: String
 
-    public init(type: String, value: String, stacktrace: [StacktraceLine], context: [String : Any], tags: [String : String], isHandled: Bool, handledBy: String) {
+    public init(type: String, message: String, stacktrace: [StacktraceLine], context: [String : Any], tags: [String : String], isHandled: Bool, handledBy: String) {
         self.type = type
-        self.value = value
+        self.message = message
         self.stacktrace = stacktrace
         self.context = context
         self.tags = tags
@@ -48,7 +48,7 @@ public class SentryEventJSException: Event {
         self.init()
         
         // Generate exception based on JavaScript exception parameters
-        let sentryException = Exception(value: jsException.value, type: jsException.type)
+        let sentryException = Exception(value: jsException.message, type: jsException.type)
         
         // Generate the stacktrace frames
         let frames = jsException.stacktrace.map {
--- a/Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
+++ b/Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
@@ -37,35 +37,47 @@ public struct JSException {
 
 public class SentryEventJSException: Event {
     override required init() {
-        // All JavaScript exceptions should be trated as fatal errors.
+        // All JavaScript exceptions should be trated as fatal errors
         super.init(level: .fatal)
-        // Setting the event's platform to JavaScript is required by Sentry to be processed as a JavaScript exception.
-        // Otherwise, Sentry won't symbolicate the stack trace.
+        // Setting the event's platform to JavaScript is required by Sentry to be processed
+        // as a JavaScript exception. Otherwise, Sentry won't symbolicate the stack trace.
         self.platform = "javascript"
     }
     
-    public static func initWithException(_ rawException: [AnyHashable: Any]) -> SentryEventJSException {
+    public static func initWithException(_ jsException: JSException) -> SentryEventJSException {
         let sentryEvent = self.init()
         
-        // Generate exception based on JavaScript exception parameters.
-        let sentryException = Exception(value: rawException["value"] as! String, type: rawException["type"] as! String)
+        // Generate exception based on JavaScript exception parameters
+        let sentryException = Exception(value: jsException.value, type: jsException.type)
         
-        // Generate the stacktrace frames.
-        var frames:[Frame] = []
-        let stacktrace = rawException["stacktrace"] as! [[AnyHashable: Any]]
-        for entry in stacktrace {
+        // Generate the stacktrace frames
+        let frames = jsException.stacktrace.map {
             let frame = Frame()
-            frame.fileName = entry["filename"] as! String
-            frame.function = entry["function"] as! String
+            frame.fileName = $0.filename
+            frame.function = $0.function
             frame.inApp = true
-            frame.lineNumber = entry["lineno"] as? NSNumber ?? 0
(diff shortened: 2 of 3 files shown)


Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,52 @@
...
+public class SentryEventJSException: Event {
+    override required init() {
+        // All JavaScript exceptions should be trated as fatal errors.
+        super.init(level: .fatal)
+        // Setting the event's platform to JavaScript is required by Sentry to be processed as a JavaScript exception.
+        // Otherwise, Sentry won't symbolicate the stack trace.
+        self.platform = "javascript"
+    }
+    
+    public static func initWithException(_ rawException: [AnyHashable: Any]) -> SentryEventJSException {
```
fluiddot: We'll define in the next iteration the structure of the raw exception object.

Sources/Remote Logging/Crash Logging/CrashLogging.swift
```diff
@@ -146,6 +146,22 @@ public class CrashLogging {
 // MARK: - Manual Error Logging
 public extension CrashLogging {
 
+    /// Writes a JavaScript exception to the Crash Logging system, including its stack trace.
+    /// Note that this function is provided mainly for hybrid sources like React Native.
+    ///
+    /// - Parameters:
+    ///   - exception: The exception object
+    ///   - callback: Callback triggered upon completion
+    func logJavaScriptException(_ jsException: JSException, callback: @escaping () -> Void) {
```
crazytonyli: The `callback` argument _feels_ redundant to me. I have asked about removing it in https://github.com/WordPress/gutenberg/pull/59221#pullrequestreview-1907518762.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,98 @@
...
+public class SentryEventJSException: Event {
+    required init() {
+        // All JavaScript exceptions should be trated as fatal errors
+        super.init(level: .fatal)
+        // Setting the event's platform to JavaScript is required by Sentry to be processed
+        // as a JavaScript exception. Otherwise, Sentry won't symbolicate the stack trace.
+        self.platform = "javascript"
+    }
+    
+    public static func initWithException(_ jsException: JSException) -> SentryEventJSException {
```
crazytonyli: Is it possible to declare this function as `public convenience init(exception: JSException)`?

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,98 @@
...
+        self.tags = tags
+        self.isHandled = isHandled
+        self.handledBy = handledBy
+    }
+    
+    public struct StacktraceLine {
+        public let filename: String?
+        public let function: String?
+        public let lineno: NSNumber?
+        public let colno: NSNumber?
```
crazytonyli: I have asked about the nullability declarations in https://github.com/WordPress/gutenberg/pull/59221#pullrequestreview-1907532299

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,98 @@
...
+import Sentry
+
+public struct JSException {
+    public let type: String
+    public let value: String
+    public let stacktrace: [StacktraceLine]
+    public let context: [String: Any]
+    public let tags: [String: String]
+    public let isHandled: Bool
+    public let handledBy: String
```
crazytonyli: Nitpick: If `handledBy` is declared as `String?`, I think the `isHandled` property can be implied as `var isHandled: Bool { handledBy != nil }`?

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,98 @@
+import Foundation
+import Sentry
+
+public struct JSException {
+    public let type: String
+    public let value: String
```
crazytonyli: Could this property be named as `message`?

Sources/Remote Logging/Crash Logging/CrashLogging.swift
fluiddot: I shared a comment about this [here](https://github.com/WordPress/gutenberg/pull/59221/files/c017e122aa58dbfcc5e33fad72a1b9207ac77880#r1507239536). In React Native, the communication between JavaScript and native is asynchronous so it's common to use callbacks.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: I'll update it based on what I shared [here](https://github.com/WordPress/gutenberg/pull/59221/files/c017e122aa58dbfcc5e33fad72a1b9207ac77880#r1507254645) 👍 .

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: Although both seem connected, I'm hesitant to infer that having `handledBy` defined implies that `isHandled` should be `false`. Let me explain a couple of cases to clarify it.

`isHandled` determines if the exception is handled by the editor, currently, this happens for exceptions captured by the error boundary components ([reference](https://github.com/WordPress/gutenberg/blob/c017e122aa58dbfcc5e33fad72a1b9207ac77880/packages/block-editor/src/components/block-list/block-crash-boundary.native.js#L20-L31)). When an error is encountered, for instance, at block level we display an error message but let the user continue editing a post. In this case, `handledBy` will tell the mechanism used to handle it and `isHandled` will be `true`.

On the other hand, other exceptions can be captured via the global error handler ([reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/f1698d65c22ce90a9552e0c6ce316c3c50d10660/src/errorLogging/index.js#L14-L43)). These are exceptions produced outside the error boundaries. Here, we use `handledBy` to tell that the exception was captured by the global error handler but we mark them as unhandled (i.e. `isHandled` as `false`) because the editor will crash.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: Sure, I used `value` to match Sentry exception parameters ([reference](https://develop.sentry.dev/sdk/event-payloads/exception/#attributes)) but I'm happy to rename it.

fluiddot: Suggestion applied in https://github.com/WordPress/gutenberg/pull/59221/commits/363cb123ba416622269c45acaf5e8df23a9151e9, https://github.com/wordpress-mobile/WordPress-iOS/pull/22655/commits/037244d34e1ec231ac66add2acfce615b0772572  and https://github.com/Automattic/Automattic-Tracks-iOS/pull/278/commits/07a7012a5af7a5e8c47bc5959e5d1331a8cffe6f.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: Suggestion applied in https://github.com/Automattic/Automattic-Tracks-iOS/pull/278/commits/ba6686962376f334881a18641515db92f0bfb4b0.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
crazytonyli: Thanks for the explanation!

Sources/Remote Logging/Crash Logging/CrashLogging.swift
jhnstn: I'm wondering about the callback too. It doesn't seem to add much value when passed this far down the call stack. As you suggest in [this comment](https://github.com/WordPress/gutenberg/pull/59221#discussion_r1507248660), it's only needed to inform about a malformed `rawException`. Even there I'm not sure how useful it will be.

fluiddot: @jhnstn As shared [here](https://github.com/WordPress/gutenberg/pull/59221#discussion_r1508682107), the main purpose of the callback is to ensure that an unhandled exception is sent to Sentry before letting the app crash. Hence, I'd advocate keeping it here.

fluiddot: @crazytonyli Regarding [this comment](https://github.com/WordPress/gutenberg/pull/59221#discussion_r1508336687), I think we need to have the callback here in Tracks because we need to guarantee that it's triggered after flushing all queued events:

https://github.com/Automattic/Automattic-Tracks-iOS/blob/07a7012a5af7a5e8c47bc5959e5d1331a8cffe6f/Sources/Remote%20Logging/Crash%20Logging/CrashLogging.swift#L159-L162

This implementation is based on other functions we have to ensure that events are sent, e.g.:
https://github.com/Automattic/Automattic-Tracks-iOS/blob/07a7012a5af7a5e8c47bc5959e5d1331a8cffe6f/Sources/Remote%20Logging/Crash%20Logging/CrashLogging.swift#L247-L274

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,96 @@
+import Foundation
+import Sentry
+
+public struct JSException {
```
crazytonyli: I wonder if it'd be nicer to change `JSException` and `StacktraceLine` to protocols. That means we can avoid the duplication between this struct implementations here and the `GutenbergJSException` in [gutenberg](https://github.com/WordPress/gutenberg/pull/59221/files#diff-879e47f147cb3b7bcda3382ba5bf1888e7208274b4dddade8582e131314603f0), by adding protocol conformations to `GutenbergJSException` at the app level.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,96 @@
...
+        self.stacktrace = stacktrace
+        self.context = context
+        self.tags = tags
+        self.isHandled = isHandled
+        self.handledBy = handledBy
+    }
+    
+    public struct StacktraceLine {
+        public let filename: String?
+        public let function: String?
```
crazytonyli: Should this `funciton` property be non-optional to match the one declared in `GutenbergJSException`?

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: That's interesting. The only problem I see is that Gutenberg and Tracks are separate libraries that are disconnected from each other. In this scenario, where should the protocols for `JSException` and `StacktraceLine` be defined?

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: Ah, good catch, I forgot to update this struct. I'll update it asap.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: My assumption in the current implementation is that WordPress-iOS needs to act as a proxy to pass the exception between libraries and hence, be in charge of converting the exception from `Gutenberg.JSException` to `Tracks.JSException`. Not sure how to conform a protocol in this case.

crazytonyli: > where should the protocols for JSException and StacktraceLine be defined?

They should be defined in Tracks. But with protocol, you can remove the initializer and definite all properties as read-only. For example.

```swift
public protocol JSException {
  var type: String { get }
  var message: String { get }
  // ...
}
```

And in the app code, only you need to do is declare protocol conformation:
```swift
import Gutenberg

extension GutenbergJSException: Tracks.JSException {
  // You should not need to write any code here, as long as the property declarations
  // in `Tracks.JSException` protocol and `GutenbergJSException` match.
}

extension GutenbergJSException.StacktraceLine: Tracks.JSException.StacktraceLine {}
```

crazytonyli: If that works, the app just passes the same object from Gutenberg to Tracks, instead of copying one struct as another.

fluiddot: Ah, I see. So, if I understand correctly, WP-iOS would expand `GutenbergJSException` struct defined in the Gutenberg Mobile library to conform to the protocol defined in Tracks 👍. I'll try this, thanks @crazytonyli for the tip 🙇 !

fluiddot: @crazytonyli I applied your suggestion in https://github.com/Automattic/Automattic-Tracks-iOS/pull/278/commits/270a150509e28eeb8e8142a0d4304ff7101a9e17 and https://github.com/wordpress-mobile/WordPress-iOS/pull/22655/commits/a1a692f09379dc1e081781d78ae488238eb956a6. The only caveat is that I had to redefine the `stacktrace` property because I encountered issues to match the protocol.

crazytonyli: Thanks for giving it a go! The issue you ran into may have something to do with Swift's type system, where it requires an exact match on the `stacktrace` property type between protocol and the struct. You should be able to "workaround" it by declaring `stacktrace` property as a generic type.

Here is an example:

```swift
// In Tracks

enum Tracks {
    protocol StacktracLine {}

    protocol JSException {
        // Key change: Add an `associatedtype` to declare the protocol type as a generic type.
        associatedtype SL = StacktracLine
        var stacktrace: [SL] { get }
    }
}

// In gutenberg

struct GutenbergeException {
    struct StacktraceLine {
    }
    var stacktrace: [StacktraceLine]
}

// In WordPress

extension GutenbergeException.StacktraceLine: Tracks.StacktracLine {}
extension GutenbergeException: Tracks.JSException {}

```

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
```diff
@@ -0,0 +1,79 @@
...
+    var type: String { get }
+    var message: String { get }
+    var jsStacktrace: [StacktraceLine] { get }
+    var context: [String: Any] { get }
+    var tags: [String: String] { get }
+    var isHandled: Bool { get }
+    var handledBy: String { get }
+}
+
+public protocol StacktraceLine {
```
crazytonyli: Too bad we can't nest protocol declarations. What do you think giving it a more specialized name like `JSStacktraceLine`?

AliSoftware: …Unless you update to Xcode `15.3` and [Swift `5.10`](https://www.swift.org/blog/swift-5.10-released/)… which [just gained such ability via SE-404](https://github.com/apple/swift-evolution/blob/main/proposals/0404-nested-protocols.md) 😉

crazytonyli: It appears you still can't nest protocol in another protocol:

```
$ swift repl
Welcome to Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4).
Type :help for assistance.
  1> protocol Foo { protocol Bar {} }
error: repl.swift:1:25: error: protocol 'Bar' cannot be nested in protocol 'Foo'
protocol Foo { protocol Bar {} }
                        ^


  1> struct Foo { protocol Bar {} }
  2>
```

fluiddot: > Too bad we can't nest protocol declarations. What do you think giving it a more specialized name like `JSStacktraceLine`?

Sure, I'll rename it to make it more specific to JavaScript.

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: That's interesting. I saw a similar workaround when reading about protocols and cases like this one. I confirmed it works in  https://github.com/Automattic/Automattic-Tracks-iOS/pull/278/commits/d3694e8a352f8cac3f3c26f73125614408c01abc 🎊, thanks @crazytonyli for the suggestion 🙇 !

Sources/Remote Logging/Crash Logging/SentryEventJSException.swift
fluiddot: > > Too bad we can't nest protocol declarations. What do you think giving it a more specialized name like `JSStacktraceLine`?
> 
> Sure, I'll rename it to make it more specific to JavaScript.

Suggestion applied in https://github.com/Automattic/Automattic-Tracks-iOS/pull/278/commits/d3694e8a352f8cac3f3c26f73125614408c01abc.

crazytonyli (approved):
LGTM!

I'm not familiar with this codebase and not sure how feasible/easy it is to add unit tests, but it'd be nice to add some unit tests to ensure the event payload scheme is correct, i.e. no `threads` and `debug_meta`.

fluiddot:
> I'm not familiar with this codebase and not sure how feasible/easy it is to add unit tests, but it'd be nice to add some unit tests to ensure the event payload scheme is correct, i.e. no `threads` and `debug_meta`.

Yep, I haven't added unit tests yet to this repository either. I briefly checked the unit tests of `CrashLogging` class and noticed that the ones related to error reporting are disabled ([reference](https://github.com/Automattic/Automattic-Tracks-iOS/blob/b6979ef69b4b094c8809ba83661222dcd0d7667e/Tests/Tests/CrashLoggingTests.swift#L97-L129)). I can take a look in a separate PR to incorporate tests for the logic around parsing a JavaScript exception.

Thank you so much @crazytonyli for your suggestions and for reviewing the PR 🙇 !

fluiddot:
I noticed that the job `buildkite/tracks-ios/validating-podspec` is failing in `trunk` ([reference](https://buildkite.com/automattic/tracks-ios/builds/455#018c83cb-18e1-45ca-bf18-c93512d6c932/6-2009)) and other PRs ([reference](https://buildkite.com/automattic/tracks-ios/builds/450#018c7ed5-ede6-4946-98fd-b434ae1c0882/6-2008)). In fact, I saw https://github.com/Automattic/Automattic-Tracks-iOS/pull/274#issuecomment-1861711772 on a past PR stating that should be safe to merge. Hence, I'm going to proceed with merging the PR.

PR #279, opened 2024-03-06
Preserve source compatibility by adding a default value.

bjhomer:
When `logError()` added a required "tags:" parameter in version 3.2.0, this broke source compatibility. Swift Package Manager uses semantic versioning, which means that upgrading from 3.x to 3.y should be source compatible. So lets add a default value so that this can remain source-compatible.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
-_None._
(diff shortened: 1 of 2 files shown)


crazytonyli (approved):

bjhomer:
Don't merge yet; I'm dealing with a potential issue.

bjhomer:
Nevermind, this PR isn't needed. I was misinterpreting an error message.

PR #281, opened 2024-03-10
Remove two deprecated status bar APIs

### LINKED ISSUE #273: Remove deprecated `statusBarFrame` usages

crazytonyli:
Fixes https://github.com/Automattic/Automattic-Tracks-iOS/issues/273.

`UIApplication.statusBarOrientation` is deprecated. Plus, the existing code compares "interface orientation" with `UIDeviceOrientation` which may be problematic.

`UIApplication.statusBarFrame` is also deprecated. I don't imagine we'd care about "status bar" height nowadays. So I have changed the implementation to always return a `0` value.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,8 @@ _None._
 
 ### Internal Changes
 
-_None._
+- The `device_info_status_bar_height` event property value now will always be zero. [#281]
+- Calculate `device_info_orientation` event property value based on "device orientation" rather than "interface orientation". [#281]
 
 ## 3.3.0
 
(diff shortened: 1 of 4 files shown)


mokagio (approved):
Tested this via the demo app and breakpoints.

***Portrait***

image


***Landscape***

image


***Unknown?!***

I got this on the very first launch and try. I tried to relaunch and reproduce but couldn't. Given our apps behavior and the data we might want to collect, I think having an occasional "Unknown" device orientation is acceptable.

image

crazytonyli:
@mokagio Thanks for the testing!

PR #283, opened 2024-03-20
Fixes macOS build

jleandroperez:
### Details:
In this PR we're fixing the macOS build.

Apologies, I didn't check #282 on the mac!!

### DIFF
--- a/Sources/Model/ObjC/Common/TracksDeviceInformation.m
+++ b/Sources/Model/ObjC/Common/TracksDeviceInformation.m
@@ -21,8 +21,10 @@
 @property (nonatomic, assign) BOOL isReachableByWiFi;
 @property (nonatomic, assign) BOOL isReachableByWWAN;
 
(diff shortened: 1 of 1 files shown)


Sources/Model/ObjC/Common/TracksDeviceInformation.m
```diff
@@ -21,8 +21,10 @@ @interface TracksDeviceInformation ()
 @property (nonatomic, assign) BOOL isReachableByWiFi;
 @property (nonatomic, assign) BOOL isReachableByWWAN;
 
+#if TARGET_OS_SIMULATOR
```
crazytonyli: Should this be `TARGET_OS_IPHONE` instead? Otherwise, these two properties won't be included when building against real device, right?

crazytonyli (approved):

jleandroperez:
Thanks a million sir!!

PR #289, opened 2024-06-13
Use Sentry 8.29.0, compatible with Xcode 16.0 Beta 1

leandroalonso:
What the title says.

---


crazytonyli (approved):
LGTM.

I don't think the apps need to wait for this change to be released though. You should be able to just update the Sentry Swift package or pod in your app project.

mokagio (approved):

PR #291, opened 2024-06-19
Expose onCrashedLastRun

1Conan:
This PR exposes Sentry's `onCrashedLastRun` option. This allows us to fetch the crash event on startup.

---

### DIFF
--- a/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
+++ b/Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Sentry
 
 #if SWIFT_PACKAGE
 import AutomatticTracksModel
@@ -17,6 +18,8 @@ public protocol CrashLoggingDataProvider {
     var enableAppHangTracking: Bool { get }
     /// Whether HTTP client errors are captured.
     var enableCaptureFailedRequests: Bool { get }
+
+    var onCrashedLastRun: ((Event) -> Void)? { get }
 }
 
 /// Default implementations of common protocol properties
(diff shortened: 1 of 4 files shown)


Sources/Remote Logging/Crash Logging/CrashLoggingDataProvider.swift
```diff
@@ -17,6 +18,8 @@ public protocol CrashLoggingDataProvider {
     var enableAppHangTracking: Bool { get }
     /// Whether HTTP client errors are captured.
     var enableCaptureFailedRequests: Bool { get }
+
+    var onCrashedLastRun: ((Event) -> Void)? { get }
```
mokagio, line 23:
```suggestion
    /// A closure that will run shortly after the crash logging is initialized in case a crash was detected during the previous run.
    var onCrashedLastRun: ((Event) -> Void)? { get }
```

For reference, the docs say, https://github.com/getsentry/sentry-cocoa/blob/1e803dcc7a1d6391e8ae84b498509fe27fe7e523/Sources/Sentry/Public/SentryOptions.h#L132-L141

```
 * A block called shortly after the initialization of the SDK when the last program execution
 * terminated with a crash.
 * @discussion This callback is only executed once during the entire run of the program to avoid
 * multiple callbacks if there are multiple crash events to send. This can happen when the program
 * terminates with a crash before the SDK can send the crash event. You can look into @c beforeSend
 * if you prefer a callback for every event.
 * @warning It is not guaranteed that this is called on the main thread.
 */
```

mokagio (approved):
Just a documentation suggestion.

Thank you for implementing this.

See also #292 which added logging for this property.

PR #310, opened 2025-12-29
Constrain Sentry version to less than 8.49 for CocoaPods

mokagio:
> [!WARNING]  
> Update: This change has been superseded by #311 

---

See https://linear.app/a8c/issue/AINFRA-1676 related to the library validation error revealed by
https://github.com/Automattic/Automattic-Tracks-iOS/pull/307: https://buildkite.com/automattic/tracks-ios/builds/567/steps/canvas?jid=019b6982-8d11-4849-bc9b-8f520a104b27#019b6982-8d11-4849-bc9b-8f520a104b27/L3836

Notice `podspec` validation passes in this PR. 

I didn't spend extra time on this right now because the issue is only just a deprecation warning _and_ the constraint affects only CocoaPods installations, which we support for the sake of it but none of our clients use.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #311, opened 2026-01-06
Remove support for distribution via CocoaPods 

mokagio:
See https://linear.app/a8c/issue/AINFRA-1693

---


mokagio:
Pausing work on this for the moment while I gather input on how to address the failing version consistency check.

Problem: We want the version hardcoded in the source and sent as part as the user agent to be in sync with the library's version as used by the package manager. This was straightforward with CocoaPods because it required hardcoding the version in the `podspec`. It is no longer as simple with SwiftPM, which delegates version declaration to Git.

There are of course possible solutions we could implement for this, like a CI check that runs only on tags, or updating the script to do something like `git fetch && git tag | sort -V | tail -1` to extract the most recent tag and compare it against the hardcoded value.

However, taking a step back I wonder whether we can do away with the library version in the user agent in the first place, thus removing the need for the validation step.

I asked about this internally, ref p1767675546597369-slack-C03DKP1JP.

CHANGELOG.md
```diff
@@ -46,7 +46,7 @@ _None._
 
 ### Internal Changes
 
-- Constrained version to `< 8.49.0` for CocoaPods installations to avoid deprecation warnings till we upgrade Sentry to version 9.x [#310]
+- With this version, Tracks only supports distribution via Swift Package Manager [#311]
```
mokagio, line 49: It seemed redundant to track the #310 workaround given it was CocoaPods related.

twstokes (approved):
I did a quick scan and didn't spot anything. This should be a low-risk change anyway, so approving. 👍

twstokes:
> However, taking a step back I wonder whether we can do away with the library version in the user agent in the first place, thus removing the need for the validation step.

Connecting dots as it looks like this was determined in https://github.com/Automattic/Automattic-Tracks-iOS/pull/313.

PR #313, opened 2026-01-07
Remove Tracks version from User Agent

mokagio:
This has been discussed internally and we decided that the overhead of keeping the version in sync between code and Git tags was not worth the value of tracking it in the user agent.

Given we already track the version of the app that logs events via the library, the library's version can always be deduced.

See https://linear.app/a8c/issue/AINFRA-1698/remove-version-from-user-agent

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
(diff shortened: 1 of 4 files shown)


mokagio:
I'm going to merge this PR into #311 without a review, I just wanted to make the change on its own for posterity.

PR #315, opened 2026-01-20
Update sentry to version 9.4

SergioEstevao:
Fixes AINFRA-1677

This PR updates Tracks to use Sentry version 9.1 

Sentry SDK 9.1 has a minimum version of iOS 15 so this also force us to update Tracks library to be minimum iOS 15 and above.

---

- [ ] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

### DIFF
--- a/Tests/Tests/CrashLoggingTests.swift
+++ b/Tests/Tests/CrashLoggingTests.swift
@@ -28,7 +28,7 @@ class CrashLoggingTests: XCTestCase {
             _ = try CrashLogging(dataProvider: mockDataProvider).start()
             XCTFail("The call above should fail")
         } catch let err {
-            XCTAssertEqual("Project ID path component of DSN is missing", err.localizedDescription)
+            XCTAssertEqual("URL scheme of DSN is missing", err.localizedDescription)
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ _None._
 
 ### Internal Changes
 
-_None._
+- The Sentry version used is now 9.4.0 [#315]
 
 ## 4.0.0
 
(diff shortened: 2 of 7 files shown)


Tests/Tests/CrashLoggingTests.swift
```diff
@@ -28,7 +28,7 @@ class CrashLoggingTests: XCTestCase {
             _ = try CrashLogging(dataProvider: mockDataProvider).start()
             XCTFail("The call above should fail")
         } catch let err {
-            XCTAssertEqual("Project ID path component of DSN is missing", err.localizedDescription)
+            XCTAssertEqual("URL scheme of DSN is missing", err.localizedDescription)
```
mokagio, line 31: I wonder if we can improve this by checking the type of the error. Maybe Sentry spits out something strongly typed that we can use instead. Just so we don't depend on a string that is more likely to change (case in point this update) than a type.

mokagio:
```diff
diff --git a/Tests/Tests/CrashLoggingTests.swift b/Tests/Tests/CrashLoggingTests.swift
index c92b084..e395834 100644
--- a/Tests/Tests/CrashLoggingTests.swift
+++ b/Tests/Tests/CrashLoggingTests.swift
@@ -27,7 +27,9 @@ class CrashLoggingTests: XCTestCase {
             mockDataProvider.sentryDSN = "--invalid DSN--"
             _ = try CrashLogging(dataProvider: mockDataProvider).start()
             XCTFail("The call above should fail")
-        } catch let err {
+        } catch let err as NSError {
+            XCTAssertEqual(err.domain, SentryErrorDomain)
+            XCTAssertEqual(err.code, 100)
             XCTAssertEqual("URL scheme of DSN is missing", err.localizedDescription)
         }
     }
```

Not sure which is better to be honest. Maybe the string, as "brittle" as it might look, is actually an easier to grasp implementation.

mokagio (approved):
Thank you for tackling this @SergioEstevao !

PR #320, opened 2026-02-24
Fix logID setter silently failing when event.extra is nil

adborbas:
The logID setter used optional chaining (`self.extra?[key] = newValue`), which is a no-op when `extra` is nil. This caused Sentry crash reports to intermittently miss the encrypted log UUID. Now the setter initializes `extra` with an empty dictionary when nil before assigning the value.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
-_None._
+- Fix `logID` setter silently failing when `event.extra` is nil, causing Sentry crash reports to intermittently miss the encrypted log UUID. [#320]
 
 ### Internal Changes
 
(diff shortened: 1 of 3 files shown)


adborbas:
Hey @mokagio, I was not sure who to assign, but the past couple of PRs were assigned to you, so I have defaulted to that. Please rotate if needed. Thanks!

mokagio (approved):
Thank you for adding tests for this, too!

adborbas:
@mokagio CI does not seem to trigger for me, are there any steps required to run it?

mokagio:
@adborbas ah! I didn't notice that, sorry. CI didn't run because the PR comes from your fork and we only allow PRs from the main repo.

I triggered it manually. Just waiting for the auto-merge.

PR #323, opened 2026-03-02
Fix UIApplication non-public API usage on watchOS

leandroalonso:
## Summary

- `TARGET_OS_IPHONE` evaluates to `1` on watchOS, causing the `UIApplication` category in `UIApplication+Extensions.{h,m}` to be compiled into Watch binaries
- `UIApplication` is a non-public class on watchOS, triggering App Store rejection error 90338
- Added `&& !TARGET_OS_WATCH` to the preprocessor guards, matching the pattern already used elsewhere in the library (e.g. `TracksDeviceInformation.m`, `ApplicationFacade.swift`, `ExPlat.swift`)

## Test plan

- [ ] Build the library targeting watchOS and confirm no `UIApplication` symbols are included
- [ ] Verify iOS builds are unaffected
- [ ] Submit a Watch app using this library to App Store Connect and confirm error 90338 no longer appears

### DIFF
--- a/Sources/Model/ObjC/Common/UIApplication+Extensions.h
+++ b/Sources/Model/ObjC/Common/UIApplication+Extensions.h
@@ -1,7 +1,7 @@
 @import Foundation;
 
 
-#if TARGET_OS_IPHONE
+#if TARGET_OS_IPHONE && !TARGET_OS_WATCH
 @import UIKit;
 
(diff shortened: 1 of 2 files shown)


mokagio (approved):
> Added `&& !TARGET_OS_WATCH` to the preprocessor guards, matching the pattern already used elsewhere in the library

Sounds good.

PR #328, opened 2026-03-26
Address flaky event logging tests

mokagio:
Recently, I noticed the Mac tests being flaky in CI. I got Codex to look into it and this seems like a decent way to address the issue.

## Summary
- stop rescheduling repeated delegate-cancelled uploads so they do not leak callbacks into later tests
- wait for `didFinishUploadingLog` in the upload manager success test before asserting completion
- keep the change scoped to the event logging flaky-test fix identified from recent Buildkite failures

## Testing

Created #329 with `100` parallel test builds. All green

image

### DIFF
--- a/Sources/Encrypted Logs/EventLogging.swift
+++ b/Sources/Encrypted Logs/EventLogging.swift
@@ -21,6 +21,7 @@ public class EventLogging {
     /// Coordinates one-at-a-time file log dequeuing and upload
     private let lock = NSLock()
     private let processingQueue = DispatchQueue(label: "event-logging-upload")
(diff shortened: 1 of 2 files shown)


twstokes (approved):
Makes sense to me. Maybe worth adding to the changelog?

mokagio:
@twstokes 

> Maybe worth adding to the changelog?

Ah, good point! This was born as a PR to fix tests, but the fix turned out to be an implementation refinement.

@copilot open a PR to add a short chanelog entry for the changes in this PR. Read the changelog over the past year to understand the style and expected level of conciseness.

PR #331, opened 2026-04-02
Track Lockdown Mode status as a device property

jkmassel:
## Summary
The WordPress iOS app doesn't work correctly under Lockdown Mode. We'll put mitigations in place, but I'd like to know how many users this affects.

Technical Details:

- Read the `LDMGlobalEnabled` UserDefaults key and include it in mutable device properties as `device_info_lockdown_mode_enabled` (`YES`/`NO`).
- The value is cached once per process lifetime since toggling Lockdown Mode requires a device restart.

## Test plan
- [ ] Verify iOS tests pass via `bundle exec fastlane ios test`
- [ ] Confirm `device_info_lockdown_mode_enabled` appears in event payloads

---


Sources/Model/ObjC/Common/TracksDeviceInformation.m
```diff
@@ -249,6 +251,14 @@ - (NSString *)preferredContentSizeCategory {
 /// - Uses `UIContentSizeCategoryIsAccessibilityCategory` method.
 /// - This will be `false` for Mac OS and watchOS.
 ///
+- (BOOL)isLockdownModeEnabled {
+    if (!self.hasReadLockdownMode) {
+        self.cachedLockdownModeEnabled = [[NSUserDefaults standardUserDefaults] boolForKey:@"LDMGlobalEnabled"];
+        self.hasReadLockdownMode = YES;
+    }
+    return self.cachedLockdownModeEnabled;
+}
```
mokagio: Damn it's frustrating when AI ignores the comments in the code...

@copilot can you push a commit to this PR to move this new method after the isAccessibilityCategory method definition, so that the documentation comment currently right on top of the new isLockdownModeEnabled method goes back in its proper location, i.e. right on top of isAccessiblityCategory ?

Sources/Model/ObjC/Common/TracksDeviceInformation.m
```diff
@@ -22,6 +22,8 @@ @interface TracksDeviceInformation ()
 @property (nonatomic, assign) BOOL isReachable;
 @property (nonatomic, assign) BOOL isReachableByWiFi;
 @property (nonatomic, assign) BOOL isReachableByWWAN;
+@property (nonatomic, assign) BOOL cachedLockdownModeEnabled;
+@property (nonatomic, assign) BOOL hasReadLockdownMode;
```
mokagio, line 31: The other values are not cached. I think it's neat to cache given the fact that a device restart is needed to change this. Worth adding a line about the change in pattern?

```suggestion

// Unlike other device properties such as network reachability which can change arbitrarily,
// Lockdown Mode requires a device restart to change, so we can cache it.
//
// See https://support.apple.com/en-us/105120
@property (nonatomic, assign) BOOL cachedLockdownModeEnabled;
@property (nonatomic, assign) BOOL hasReadLockdownMode;
```

CHANGELOG.md
```diff
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Track whether Lockdown Mode is enabled as a device property (`device_info_lockdown_mode_enabled`).
```
mokagio:
```suggestion
- Track whether Lockdown Mode is enabled as a device property (`device_info_lockdown_mode_enabled`). [#331]
```

Sources/Model/ObjC/Common/TracksDeviceInformation.m
mokagio: 💙

Sources/Event Logging/TracksService.m
```diff
@@ -422,6 +423,7 @@ - (NSDictionary *)mutableDeviceProperties
              DeviceInfoOrientation : self.deviceInformation.orientation ?: @"Unknown",
              DeviceInfoPreferredContentSizeCategory : self.deviceInformation.preferredContentSizeCategory ?: @"Unknown",
              DeviceInfoIsAccessibilityCategory : self.deviceInformation.isAccessibilityCategory ? @"YES" : @"NO",
+             DeviceInfoLockdownModeEnabledKey : self.deviceInformation.isLockdownModeEnabled ? @"YES" : @"NO",
```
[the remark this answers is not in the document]
mokagio: Given this is unrelated to this PR, @copilot create a new PR to remove this redundancy

Sources/Event Logging/TracksService.m
```diff
@@ -66,6 +66,7 @@ @interface TracksService ()
 NSString *const DeviceInfoOrientation = @"device_info_orientation";
 NSString *const DeviceInfoPreferredContentSizeCategory = @"device_info_preferred_content_size_category";
 NSString *const DeviceInfoIsAccessibilityCategory = @"device_info_is_accessibility_category";
+NSString *const DeviceInfoLockdownModeEnabledKey = @"device_info_lockdown_mode_enabled";
```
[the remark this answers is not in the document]
mokagio: @jkmassel up to you what you prefer? I'd go for the current suggestion, keeping in mind that we'd have to update usages, too, which \copilot could do for us as well

mokagio (approved):
Approving to unblock. What an interesting find this was, well done.

It's sad there's no test for this, likely due to pre-existing lack of tests. At the same time, the logic is bound to user defaults, so a test would basically be testing a thin, straightforward wrapper. FWIW, I'm okay to trade the lack of coverage for not having to put in place a harness for this. The ROI seems small.

PR #336, opened 2026-04-20
Add support for tvOS target

SergioEstevao:
---
This PR adds support for the tvOS platform type in order to be used in the future Pocket Casts tvOs app.

Besides adding the necessary platform support and compiler checks it adds a demo app in order to test the sending of the events like the other platforms.

- [ ] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.


TracksDemo/TracksDemo tvOS/TracksDemo_tvOSApp.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  TracksDemo_tvOSApp.swift
+//  TracksDemo tvOS
+//
+//  Created by Sérgio Estêvão on 20/04/2026.
+//  Copyright © 2026 Automattic Inc. All rights reserved.
+//
+
```
mokagio:
```suggestion
```

Package.swift
```diff
@@ -4,7 +4,7 @@ import PackageDescription
 
 let package = Package(
     name: "AutomatticTracksiOS",
-    platforms: [.macOS(.v10_14), .iOS(.v15)],
+    platforms: [.macOS(.v10_14), .iOS(.v15), .tvOS(.v15)],
```
mokagio, line 7: Is `v15` due to Sentry or other dependencies? It's quite old

Image

https://developer.apple.com/documentation/packagedescription/supportedplatform/tvosversion

Having said that, so are the macOS and iOS versions, so maybe we can roll with it and update in a dedicated follow up for all of them.

mokagio (approved):
Took me a while to find the events from the demo app because:

1. I had forgotten that the app by default doesn't actually send events (`shouldSendEvents = false`)
2. I had forgotten Tracks Live View will not show events that are not registered unless you ask it to

But once I figured that out, I was able to see the events. All good.

Package.swift
SergioEstevao: I just kept it inline with the other platforms. To be fair if it's compatible with that version I don't see a reason to push to a more recent version.
Regarding Sentry, will we need to change something to make it use a specific tvOs version of it?

SergioEstevao:
@mokagio do you understand why it fails the build-and-test-ios step?

mokagio:
@SergioEstevao 

> @mokagio do you understand why it fails the build-and-test-ios step?

I came here to comment on that, actually. I looked into that yesterday (and I should have left a comment about it, sorry).

Long story short, #338 fixes the issue and targets this PR. I enabled auto-merge, so once that's merged in here, this will land on `trunk` and we'll be able to publish a new release.

mokagio (approved):
Thanks again for working on this @SergioEstevao

PR #338, opened 2026-04-23
Use Xcode 26.4.1 in CI

mokagio:
When I approved #336 , I noticed the test failure. At first, I thought it was a flaky test but upon further investigation it turned out to be due to the tests for the demo app sneaking into the test plan run in CI for iOS.

It unnecessary to run those tests, given that: 1) the demo app is for internal dev only and 2) they are template only.

This PR restores the iOS test plan to how it was prior, and updates the tooling to latest Xcode 26.4.1 along the way.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. - N.A.


SergioEstevao (approved):
All good

PR #359, opened 2026-06-03
Fix release build on tvOS

kean:
Fixes build failure on release builds https://buildkite.com/automattic/pocket-casts-ios/builds/15417#019e8ec5-9bac-42e4-bc8f-dc0d37e9fd56.


SergioEstevao (approved):
Good catch

PR #361, opened 2026-06-11
Store Tracks Core Data DB in Caches on tvOS

kean:
### Problem

`TracksContextManager` creates its Core Data SQLite store under **Application Support**, which isn't writable on tvOS (only Caches and tmp are). On a real Apple TV, `TracksService` init throws `TracksApplicationSupportException` / `TracksPersistentStoreException`. The tvOS *Simulator* hides this (it inherits the Mac filesystem), so it only fails on device.

### Fix

Route the store's base directory to `NSCachesDirectory` on tvOS, keeping Application Support on every other platform. The store becomes `Library/Caches/<bundleid>/Tracks.sqlite` on tvOS. Caches is purgeable, which is fine for the transient Tracks event queue (events are flushed to the server). Also made the directory-creation error text platform-neutral and the store-directory test platform-aware.

### Testing

`testStoreIsInExpectedDirectory` passes on tvOS (store under Caches) and macOS (Application Support); `build-for-testing` succeeds for the tvOS Simulator.

---

### DIFF
--- a/Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
+++ b/Sources/Model/ObjC/Common/Core Data/TracksContextManager.m
@@ -92,28 +92,38 @@ NSString *const TracksPersistentStoreException      = @"TracksPersistentStoreExc
                                               attributes:nil
                                                    error:&error];
     
-    // It seems safe not to handle this error because Application Support should always be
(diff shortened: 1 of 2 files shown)


SergioEstevao (approved):
:shipit:

PR #364, opened 2026-06-23
Make ExPlat.shared thread-safe

adborbas:
## Description

`ExPlat.shared` was a process-wide mutable global declared as a plain stored static var. The SDK **reassigns and releases** it on a background thread — every `ExPlat.configure(…)` builds a new instance whose `init` ends with `ExPlat.shared = self`, and consumers reach `configure` from background queues via `TracksService` user-switches — while consumers read `ExPlat.shared` and call `refresh` on the main thread, with no synchronization.

Because a *stored* static var hands call sites the value at +0 (no retain across the call), a reader can borrow an instance that a background `configure` then releases to refcount 0. The main-thread `refresh` forms `[weak self]` (`objc_initWeak`) on the dying instance and aborts:

```
SIGABRT: Cannot form weak reference to instance … of class AutomatticExperiments.ExPlat.
```

## Fix

Replace the stored static var with a lock-guarded computed property backed by private storage (`NSLock` + `_shared`). Two effects fix the crash:

1. A computed getter returns an **owned (+1)** value, so `ExPlat.shared?.refresh()` retains the instance for the duration of the call — eliminating the +0 borrow that allowed mid-`refresh` deallocation. No consumer code changes required.
2. The lock makes get/set atomic, so a reader never observes a torn/half-reassigned pointer and the release of the previous instance can't interleave with a read.

`NSLock` with explicit `lock()`/`defer unlock()` is used because the deployment targets (macOS 10.14 / iOS 15 / tvOS 15) predate `OSAllocatedUnfairLock` and `NSLock.withLock`. The public API and the `@objc` `configureWithPlatform:…` entry point are unchanged.

## Testing

- Added `ExPlatConcurrencyTests` (Swift Testing): repeatedly reassigns `ExPlat.shared` on a background queue while concurrently calling `refresh`/`experiment` on another, for 1000 iterations.
- Verified the test **reports a data race and fails** against the pre-fix stored var under Thread Sanitizer, and is **TSan-clean and passes** with the fix.

---

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -20,6 +20,14 @@ steps:
     plugins: [$CI_TOOLKIT_PLUGIN]
     artifact_paths: .build/logs/*.log
 
+  - label: "🧪 ExPlat Thread Sanitizer"
+    key: test_explat_tsan
+    command: |
+      install_gems
+      bundle exec fastlane ios test_explat_concurrency_tsan clean:true
+    plugins: [$CI_TOOLKIT_PLUGIN]
+    artifact_paths: .build/logs/*.log
+
   ######################
   # Build and Test macOS
   ######################
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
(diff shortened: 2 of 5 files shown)


mokagio (approved):
Thank you for working on this!

I pushed a few commits to refine. 

In particular, I didn't noticed the line about TSan in the PR description:

> Verified the test reports a data race and fails against the pre-fix stored var under Thread Sanitizer, and is TSan-clean and passes with the fix.

So, I was surprised when the tests passed on my end with the locking code disabled. I then realized the issue and verified the crash with `-enableThreadSanitizer YES`.

